PR #63, opened 2015-01-29
Allow CAR deployment with sequence name 'main'/'fault'

kishokpriya:
Fix for ESBJAVA-3147


chanakaudaya:
This is already merged to the code with a different PR sent to the develop branch when we were using develop branch.

PR #75, opened 2015-02-08
ESBJAVA-3110

maheeka:


shafreenAnfar:
It seems we can replace the two if conditions with the below condition. 

if (mediator == null)

maheeka:
fixed condition at https://github.com/maheeka/wso2-synapse/commit/fc0e4deec22517573b46f5e835ea0e3e3bcfcfb3

PR #79, opened 2015-02-12
Mvn dependency version fix

maheeka:
refactor all versions of dependencies and plugins to parent pom from child poms


ramith:
This might be a useful thing to do : http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

maheeka:
@ramith thanks for pointing out. Will do so in future.

PR #80, opened 2015-02-15
adding axis2 handler for build message id ws addressing validation used

isururanawaka:


Buddhima:
To use this need to add entry in PR : https://github.com/wso2/product-esb/pull/66
And in the proxy which requires WS-Addressing add the following service parameter. 

``` <parameter name="enforceWSAddressing">true</parameter>  ```

isururanawaka:
Refer https://github.com/wso2/wso2-synapse/pull/88

PR #107, opened 2015-03-20
RabbitMQ_MSMP : ( after initial code review )

ecr666:


modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitmqProducer.java
```diff
@@ -0,0 +1,174 @@
...
+			ByteArrayOutputStream os = new ByteArrayOutputStream();
+			ObjectOutput objOut = new ObjectOutputStream(os);
+			objOut.writeObject(message);
+			byte[] byteForm = os.toByteArray();
+			objOut.close();
+			os.close();
+			//building AMQP message
+			AMQP.BasicProperties.Builder builder = new AMQP.BasicProperties().builder();
+			builder.messageId(synCtx.getMessageID());
+			builder.deliveryMode(2);
```
iuj: Delivery Mode 2 == Persists the message

PR #137, opened 2015-04-08
Xml comments support in synapse configuration

jagathsisira:
With this updated, selected XML configurations will preserve XML comments in ESB configuration files.


modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXMLConfigurationFactory.java
```diff
@@ -99,11 +100,11 @@ public SynapseConfiguration getConfiguration(OMElement definitions, Properties p
                     defineEventSource(config, elt, properties);
                 } else if (XMLConfigConstants.EXECUTOR_ELT.equals(elt.getQName())) {
                     defineExecutor(config, elt, properties);
-                } else if(XMLConfigConstants.MESSAGE_STORE_ELT.equals(elt.getQName())) {
+                } else if (XMLConfigConstants.MESSAGE_STORE_ELT.equals(elt.getQName())) {
                     defineMessageStore(config, elt, properties);
```
isudana: It's better if we do not include unrelated code changes like formatting etc. to this PR.

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
```diff
@@ -2127,4 +2132,25 @@ private String getTenantDomain(SynapseEnvironment synapseEnvironment) {
         return null;
     }
 
+    /**
+     * Returns CommentedText List
+     *
+     * @return ArrayList of String contains commented text segments
+     */
+    public List<String> getCommentedTextList() {
```
prabathabey: Maybe it's a good idea not to expose internals such as the data structures used internally, etc as part of "method names".

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
```diff
@@ -2127,4 +2132,25 @@ private String getTenantDomain(SynapseEnvironment synapseEnvironment) {
...
+        return commentedTextList;
+    }
+
+    /**
+     * Add new comment text entry to the existing list*
+     *
+     * @param comment String comment
+     */
+    public void addToCommentedTextList(String comment) {
```
prabathabey: Same goes here as well.

modules/core/src/main/java/org/apache/synapse/config/xml/AbstractListMediatorFactory.java
```diff
@@ -34,18 +37,45 @@
...
+    }
+
+    /**
+     * Find and add all comment nodes to the parent mediator as Comment Mediators
+     *
+     * @param el OMElement to extract OMComment Nodes
+     * @param m  Mediator to be updated with extracted OMComment nodes
+     */
+    protected static void addAllCommentChildrenToMediator(OMElement el, ListMediator m) {
+        Iterator it = el.getChildren();
```
prabathabey: This particular iterator's scope is not properly restricted. Imagine someone using the same iterator instance somewhere down the code. If the same is modified within the while clause, then that can lead to inconsistent behaviours.

modules/core/src/main/java/org/apache/synapse/config/xml/AbstractMediatorFactory.java
```diff
@@ -212,4 +215,24 @@ protected void handleException(String message) {
...
     }
+
+    /**
+     * Find and add all comment nodes to the mediator as a list
+     *
+     * @param el          OMElement to extract OMComment Nodes
+     * @param commentList ArrayList to be updated with extracted Comment Strings
+     */
+    protected static void addAllCommentChildrenToList(OMElement el, List<String> commentList) {
+        Iterator it = el.getChildren();
```
prabathabey: Same goes here as well.

modules/core/src/main/java/org/apache/synapse/config/xml/CommentMediatorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.xml.namespace.QName;
+import java.util.Properties;
+
+/**
+ * Factory for Comment Mediators.
+ */
+public class CommentMediatorFactory extends AbstractMediatorFactory {
+
+    private static final QName Comment_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "comment");
```
prabathabey: Variable naming conventions are not properly followed.

PR #142, opened 2015-04-17
Fix for incorrect timeout handling in nhttp

VijithaEkanayake:


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/ClientHandler.java
```diff
@@ -504,8 +510,12 @@ private String getErrorMessage(String message, NHttpClientConnection conn) {
...
         final int errorCode, final String errorMessage, final Exception exceptionToRaise) {
 
-        Axis2HttpRequest axis2Request = (Axis2HttpRequest)
-                conn.getContext().getAttribute(AXIS2_HTTP_REQUEST);
+        Axis2HttpRequest axis2Request = (Axis2HttpRequest) conn.getContext().getAttribute(AXIS2_HTTP_REQUEST);
+
+        if (axis2Request == null) {
+            axis2Request = (Axis2HttpRequest) conn.getContext().getAttribute(ATTACHMENT_KEY);
+        }
+
```
Buddhima: It would be better if this is implemented as reusable manner

VijithaEkanayake: Thanks a lot for the suggestion. In following PR https://github.com/wso2/wso2-synapse/pull/361 modifications are made as the comment.

PR #169, opened 2015-04-29
ESBJAVA 3380

isururanawaka:


bsenduran:
fixed in https://github.com/wso2/wso2-synapse/pull/262

PR #177, opened 2015-05-07
Fix ESBJAVA-3663

malakasilva:


bsenduran:
correct JIRA id: https://wso2.org/jira/browse/ESBJAVA-3671

PR #182, opened 2015-05-08
Fixed for ESB-JAVA 2834

ashalya:


modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -224,9 +224,13 @@ public String getSeparator() {
     }
 
     public void setSeparator(String separator) {
-        this.separator = separator;
+        if (separator.contains("\\n")) {
+            this.separator = "\n";
```
vanjikumaran: Shouldn't it be a  (separator.equals("\\n")) ?

PR #191, opened 2015-05-13
Foreach Mediator Implementation

maheeka:


modules/core/src/main/java/org/apache/synapse/config/xml/ForEachMediatorFactory.java
```diff
@@ -0,0 +1,130 @@
...
+                    break;
+                } else if (m instanceof SendMediator) {
+                    valid = false;
+                    break;
+                }
+            }
+        }
+        return valid;
+    }
+
```
isudana: Can be simplify like follows.
private boolean validateSequence(SequenceMediator sequence) {
        if (sequence != null) {
            List<Mediator> mediators = sequence.getList();
            for (Mediator m : mediators) {
                if (m instanceof CallMediator || m instanceof CalloutMediator ||
                    m instanceof SendMediator) {
                    return false;
                }
            }
        }
        return true;
    }

isudana:
Refactored this implementation and merge with following PR
https://github.com/wso2/wso2-synapse/pull/197

modules/core/src/test/java/org/apache/synapse/mediators/builtin/ForEachHelperMediator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private List mediatedContext = new ArrayList();
+    private int msgCount;
+
+    public boolean mediate(MessageContext synCtx) {
+
+        try {
+            mediatedContext.add(MessageHelper.cloneMessageContext(synCtx));
+        } catch (AxisFault e) {
+            e.printStackTrace();
```
sandamal: Can use a logger for this

ramith:
git rebase -i HEAD~40 would have been a good thing to do ( so that commit history of mahika's private repo wouldn't end up in public synapse repository.

PR #195, opened 2015-05-14
MSMP resources are cleaned up in cluster mode.

ravindraranwala:


ravindraranwala:
code was cleaned up https://github.com/wso2/wso2-synapse/pull/209

modules/core/src/main/java/org/apache/synapse/message/processor/MessageProcessorCleanupService.java
```diff
@@ -0,0 +1,9 @@
+package org.apache.synapse.message.processor;
+
+import java.io.Serializable;
+import java.util.concurrent.Callable;
+
+public interface MessageProcessorCleanupService extends Callable<Void>, Serializable {
+    void setName(final String name);
+
```
shafreenAnfar: Remove this additional space.

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -108,7 +115,8 @@ public void init(SynapseEnvironment se) {
 
 	}
```
shafreenAnfar: It seems like we have to fix indentation.

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -388,18 +415,49 @@ private BlockingMsgSender initMessageSender(Map<String, Object> params) {
...
+            for (MessageConsumer messageConsumer : messageConsumers) {
+                messageConsumer.cleanup();
+            }
+        }
+    }
+
+    private Callable<Void> getMessageProcessorCleanupTask() {
+        MessageProcessorCleanupService cleanupTask = null;
+        if (logger.isDebugEnabled()) {
+            logger.debug("Trying to fetch InboundRequestProcessor from classpath.. ");
```
shafreenAnfar: Comment should have either one "." or three

ravindraranwala:
These issues were solved by the pull request [1] 

[1] https://github.com/wso2/wso2-synapse/pull/242

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -32,6 +36,8 @@
 import org.apache.synapse.task.TaskManager;
 import org.apache.synapse.task.TaskManagerObserver;
 
+import sun.misc.Service;
```
prabathabey: Avoid using low level private APIs such ash sun.misc.Service, which can lead to interoperability issues if the same code is run on a JDK introduced by a different vendor. The same API was moved into a public API in all new JDK versions, namely, ServiceLoader, which can effectively be used as an alternative.

ravindraranwala: This issue was fixed in https://github.com/wso2/wso2-synapse/pull/242

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
ravindraranwala: This issue was fixed in https://github.com/wso2/wso2-synapse/pull/253

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
ravindraranwala: This issue was fixed in https://github.com/wso2/wso2-synapse/pull/253

modules/core/src/main/java/org/apache/synapse/message/processor/MessageProcessorCleanupService.java
ravindraranwala: This issue was fixed in https://github.com/wso2/wso2-synapse/pull/253

PR #207, opened 2015-05-26
Fixing ESBJAVA-3384

Shakila:


malakasilva:
Reverted the fix with https://github.com/wso2/wso2-synapse/pull/207

malakasilva:
Reverted the fix with https://github.com/wso2/wso2-synapse/pull/208

PR #212, opened 2015-05-27
Fixing ESBJAVA-3384

Shakila:


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/PassThroughTransportUtils.java
```diff
@@ -169,17 +169,24 @@ public static int determineHttpStatusCode(MessageContext msgContext) {
...
-            // shall faults be transmitted with HTTP 200
-            boolean faultsAsHttp200 =
-                PassThroughConstants.TRUE.equals(
-                    msgContext.getProperty(PassThroughConstants.FAULTS_AS_HTTP_200).toString().toUpperCase());
-
-            // Set HTTP status code to 500 if this is a fault case and we shall not use HTTP 200
-            if (handleFault && !faultsAsHttp200) {
+            try {
+                // shall faults be transmitted with HTTP 200
+                boolean faultsAsHttp200 =
```
malakasilva: Still this line can through nullapointer exception. Need to check all scenarios before sending pull request

PR #214, opened 2015-06-03
Persistent Property Implementation

nadeeshaan:


nadeeshaan:
This PR is related with https://github.com/wso2/carbon-mediation/pull/198

modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
```diff
@@ -162,6 +166,33 @@ public boolean mediate(MessageContext synCtx) {
...
+
+                if (value != null){
+                    content.append(value.toString());
+                }
+                else {
+                    content.append(resultValue);
+                }
+
+                Registry registry = synCtx.getConfiguration().getRegistry();
+                registry.newNonEmptyResource(path.toString(), false, "text/plain", resultValue.toString(), propertyName.toString());
```
jagathsisira: text/plain can be defined as a constant within the class and reuse it.

PR #185, opened 2015-05-11
Fixed for ESBJAVA 3670,ESBJAVA 3686, ESBJAVA 3677

prabathariyaratna:
Fix for the message duplication after a transient outage of the endpoint and the message store.

PR #217, opened 2015-06-08
fix for ESBJAVA-3670

isudana:


isudana:
Initial fix was done by PrabathAR in https://github.com/wso2/wso2-synapse/pull/185

PR #225, opened 2015-06-10
fixing HTTP HEAD method Content-Length (0) isuue

VijithaEkanayake:


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceResponse.java
```diff
@@ -168,6 +168,23 @@ public void start(NHttpServerConnection conn) throws IOException, HttpException
...
         
         sourceConfiguration.getHttpProcessor().process(response, conn.getContext());
+
+        //Since HTTP HEAD request doesn't contain message body, entity will be null when above HttpProcessor()
+        // process invoked. Hence content length is set to 0 inside the HTTP-Core. To avoid that content length of
+        // the backend response is set as the content length.
+        if (entity == null &&
+            PassThroughConstants.HTTP_HEAD.equalsIgnoreCase(request.getRequest().getRequestLine().getMethod())) {
+            if (response.getFirstHeader(PassThroughConstants.ORGINAL_CONTEN_LENGTH) == null && (response
+                    .getFirstHeader(HTTP.CONTENT_LEN).getValue().equals("0"))) {
```
jagathsisira: Isn't it required to do a null check here getFirstHeader(HTTP.CONTENT_LEN).getValue()

VijithaEkanayake: Thanks lot for raising this. If response codes are equals to 204,205,304 Http-core doesn't set header value to 0. Hence adding null there. Following PR https://github.com/wso2/wso2-synapse/pull/362 contains the changes.

PR #238, opened 2015-06-23
Add Axis2 module engaging feature to proxy service

prabathariyaratna:


prabathariyaratna:
Jira reference https://wso2.org/jira/browse/ESBJAVA-4221

modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
```diff
@@ -720,6 +722,32 @@ public AxisService buildAxisService(SynapseConfiguration synCfg, AxisConfigurati
...
+                            proxyService.engageModule(axisModule, axisCfg);
+                            moduleEngaged = true;
+                        }
+
+                    } catch (AxisFault axisFault) {
+                        handleException("Error loading " + moduleName + " module on proxy service : "
+                                        + name, axisFault);
+                    }
+                }
+            }
```
shafreenAnfar: Remove the space.

prabathariyaratna: Fixed in https://github.com/wso2/wso2-synapse/pull/373

PR #244, opened 2015-07-01
XPath Base64 decode function

prabathariyaratna:


prabathariyaratna:
Jira reference https://wso2.org/jira/browse/ESBJAVA-4220

modules/core/src/main/java/org/apache/synapse/util/xpath/Base64DecodeFunction.java
```diff
@@ -0,0 +1,113 @@
...
+            // get the first argument, it can be a function returning a string as well
+            String encodedValue = StringFunction.evaluate(args.get(0), context.getNavigator());
+
+            // charset is in the second argument
+            String charset = StringFunction.evaluate(args.get(1), context.getNavigator());
+
+            return decode(debugOn, charset, encodedValue);
+        } else {
+            if (debugOn) {
+                log.debug("base64Decode function expects only two arguments maximum, returning empty string");
```
shafreenAnfar: Shouldn't this be an error log ?

modules/core/src/main/java/org/apache/synapse/util/xpath/Base64DecodeFunction.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final Log log = LogFactory.getLog(Base64DecodeFunction.class);
+
+    public static final String NULL_STRING = "";
+
+    private static final String DEFAULT_CHARSET = "UTF-8";
+
+    public Object call(Context context, List args) throws FunctionCallException {
+
+        boolean debugOn = log.isDebugEnabled();
+
```
shafreenAnfar: What is the purpose of setting it to a variable and then using it ?

modules/core/src/main/java/org/apache/synapse/util/xpath/Base64DecodeFunction.java
prabathariyaratna: We are not stopping the execution of the process if this happens. That's why giving an warning and continue the execution.

modules/core/src/main/java/org/apache/synapse/util/xpath/Base64DecodeFunction.java
prabathariyaratna: Fixed in https://github.com/wso2/wso2-synapse/pull/373

PR #247, opened 2015-07-08
ESBJAVA-3751 uri-template fixes

djkevincr:
added fix with unit tests.


modules/commons/src/test/java/org/apache/synapse/commons/templates/uri/parser/ParserTest.java
```diff
@@ -64,7 +62,7 @@ public void testSimpleStringExpansion() throws Exception {
 
     public void testSimpleStringMatch() throws Exception {
         URITemplate template = new URITemplate("/admin/~{user}");
-        Map<String,String> var = new HashMap<String,String>();
+        Map<String, String> var = new HashMap<String, String>();
```
prabathabey: Use meaningful names for variables!

PR #252, opened 2015-07-10
moving staxon dependency to  synapse

djkevincr:


modules/commons/src/main/java/org/apache/synapse/commons/json/JsonDataSourceImpl.java
```diff
@@ -109,7 +112,7 @@ public JsonDataSourceImpl(InputStream inputStream) {
             }
         } catch (IOException e) {
             logger.error("Could not create a JSON data source from the input stream. "
-                         + e.getLocalizedMessage());
+                    + e.getLocalizedMessage());
```
imesh: What would be the reason for using e. getLocalizedMessage() instead of e.getMessage()? Do we support translations?

modules/commons/src/main/java/org/apache/synapse/commons/json/XMLEventWriterDelegate.java
```diff
@@ -43,14 +43,15 @@ public XMLEventWriterDelegate(XMLStreamWriter delegate) {
         super(delegate);
         this.delegate = delegate;
     }
+
     public void add(XMLEvent event) throws XMLStreamException {
```
imesh: It might be better to add method comments.

modules/commons/src/main/java/org/apache/synapse/commons/staxon/core/base/AbstractXMLInputFactory.java
```diff
@@ -0,0 +1,260 @@
...
+
+    private boolean coalescing;
+    private boolean namespaceAware;
+    private boolean replacingEntityReferences;
+    private boolean supportingExternalEntities;
+    private boolean validating;
+    private boolean supportDTD;
+
+    @Override
+    public XMLStreamReader createXMLStreamReader(Source source) throws XMLStreamException {
```
imesh: It might be better to add method comments at least for the public methods.

modules/commons/src/main/java/org/apache/synapse/commons/staxon/core/base/AbstractXMLInputFactory.java
```diff
@@ -0,0 +1,260 @@
...
+            if (reader != null) {
+                if (streamSource.getSystemId() != null) {
+                    return createXMLStreamReader(streamSource.getSystemId(), reader);
+                } else {
+                    return createXMLStreamReader(reader);
+                }
+            }
+            if (streamSource.getSystemId() != null) {
+                try {
+                    final InputStream stream = new URI(source.getSystemId()).toURL().openStream();
```
imesh: Does the caller of this method need to close the stream after using it? If so might be better to mention it in the method comment.

PR #261, opened 2015-07-16
Changes for supporting ctx in payloadFactory

keerthu:


modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -132,7 +133,9 @@ private boolean mediate(MessageContext synCtx, String format) {
         }
         org.apache.axis2.context.MessageContext axis2MessageContext = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
         StringBuffer result = new StringBuffer();
-        regexTransform(result, synCtx, format);
+	    StringBuffer resultCTX = new StringBuffer();
```
malakasilva: Can't we use result (StringBuffer)

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -142,7 +145,7 @@ private boolean mediate(MessageContext synCtx, String format) {
                 JsonUtil.removeJsonPayload(axis2MessageContext);
                 OMElement omXML = AXIOMUtil.stringToOM(out);
                 if (!checkAndReplaceEnvelope(omXML, synCtx)) { // check if the target of the PF 'format' is the entire SOAP envelop, not just the body.
-                    axis2MessageContext.getEnvelope().getBody().addChild(omXML.getFirstElement());
+	                axis2MessageContext.getEnvelope().getBody().addChild(omXML.getFirstElement().getFirstElement());
```
malakasilva: Can't understand this logic. Why do you need this?

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
keerthu: If same variable is used to do the replacing in both (ctx and args) it will append the both results in the payload.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -72,8 +72,9 @@
 
     private List<Argument> pathArgumentList = new ArrayList<Argument>();
     private Pattern pattern = Pattern.compile("\\$(\\d)+");
+	private Pattern ctxPattern = Pattern.compile("\\$(ctx.[^,\"'<>\n}\\]]*)");
 
-    private static final Log log = LogFactory.getLog(PayloadFactoryMediator.class);
+	private static final Log log = LogFactory.getLog(PayloadFactoryMediator.class);
```
malakasilva: Formatting issue

keerthu: Fixed in the PR https://github.com/wso2/wso2-synapse/pull/363/files

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -72,8 +72,9 @@
 
     private List<Argument> pathArgumentList = new ArrayList<Argument>();
     private Pattern pattern = Pattern.compile("\\$(\\d)+");
+	private Pattern ctxPattern = Pattern.compile("\\$(ctx.[^,\"'<>\n}\\]]*)");
```
prabathabey: You should probably make this variable "static" to make things more memory efficient.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -159,35 +162,90 @@ private boolean mediate(MessageContext synCtx, String format) {
...
-        }
-    }
+	/**
+	 * Calls the replaceCTX function. isFormatDynamic check is used to remove indentations which come from registry based
+	 * configurations.
+	 * @param resultCTX
+	 * @param synCtx
+	 * @param format
+	 */
+	private void regexTransformCTX(StringBuffer resultCTX, MessageContext synCtx, String format) {
```
prabathabey: Be consistent on variable names etc. See "resultCTX" and "synCtx"

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -159,35 +162,90 @@ private boolean mediate(MessageContext synCtx, String format) {
...
+		if (isFormatDynamic()) {
+			String key = formatKey.evaluateValue(synCtx);
+			Object entry = synCtx.getEntry(key);
+			if(entry == null){
```
prabathabey: Formatting issues.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -159,35 +162,90 @@ private boolean mediate(MessageContext synCtx, String format) {
...
+				handleException("Key " + key + " not found ", synCtx);
+			}
+			String text = "";
```
prabathabey: Any reason not to null-init this variable?

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -159,35 +162,90 @@ private boolean mediate(MessageContext synCtx, String format) {
...
+	 * Replaces the payload format with property values from messageContext.
+	 *
+	 * @param format
+	 * @param resultCTX
+	 * @param synCtx
+	 */
+	private void replaceCTX(String format, StringBuffer resultCTX, MessageContext synCtx) {
+		Matcher ctxMatcher;
+
+		if (mediaType != null && (mediaType.equals(JSON_TYPE) || mediaType.equals(TEXT_TYPE))) {
```
prabathabey: Could have simply avoided this null check had we compared "mediaType" against the known constants.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -284,6 +338,19 @@ private String inferReplacementType(Map.Entry<String, String> entry) {
         }
     }
 
+	private String inferReplacementType(String entry) {
+		if(isWellFormedXML(entry)) {
+			return XML_TYPE;
```
prabathabey: Why not use an enum?

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -159,35 +162,90 @@ private boolean mediate(MessageContext synCtx, String format) {
...
+		} else {
+			ctxMatcher=ctxPattern.matcher("<pfPadding>" + format + "</pfPadding>");
+		}
+		while (ctxMatcher.find()) {
+			String ctxMatchSeq = ctxMatcher.group();
+			String expressionTxt = ctxMatchSeq.substring(5, ctxMatchSeq.length());
+
+			String replaceValue = synCtx.getProperty(expressionTxt).toString();
+
+			if(mediaType.equals(JSON_TYPE) && inferReplacementType(replaceValue).equals(XML_TYPE)) {
```
nuwand: Could have been better to have as if(JSON_TYPE.equals(mediaType)) to avoid possible NPEs. Also check formatting. Check whether your IDE is configured with the wso2 code template.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -132,7 +133,9 @@ private boolean mediate(MessageContext synCtx, String format) {
         }
         org.apache.axis2.context.MessageContext axis2MessageContext = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
         StringBuffer result = new StringBuffer();
```
lafernando: StringBuilder should be used instead of StringBuffer, if multiple threads are not using it at the same time.

PR #277, opened 2015-07-21
fixing integration test issues throttle mediator

djkevincr:


modules/extensions/src/main/java/org/apache/synapse/mediators/throttle/ThrottleMediator.java
```diff
@@ -113,9 +113,8 @@ public boolean mediate(MessageContext synCtx) {
             //we consider dynamic loading of policy loading only for the request flow mediation
             //we ignore policy initialization for the response flow case we use the existing policy
             //reference throttling only applies for request flow mediation
-            if (!isResponse) {
```
chanakaudaya: What is the reason for removing this check?

djkevincr: This check is redundant since is it is enclosed by the same check above.

PR #280, opened 2015-07-23
Fix for: https://wso2.org/jira/browse/ESBJAVA-3675

virajsenevirathne:
Fixed Issue:
When Iterate Mediator is used with  sequential set to true, if fault occurs inside the iteration sequence ESB won't able to execute the onError fault sequence of that as that fault stack is lost with cloned iterated message

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
@@ -166,7 +166,7 @@ public class IterateMediator extends AbstractMediator implements ManagedLifecycl
                         throw synEx;
                     } catch (Exception e) {
                         copyFaultyIteratedMessage(synCtx, iteratedMsgCtx);
-                        handleException("Exception Occurred while executing sequential iteration " +
+                        handleException("Exception occurred while executing sequential iteration " +
                                         "in the Iterator Mediator", e, synCtx);
                     }
                 }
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
```diff
@@ -142,11 +146,29 @@ public boolean mediate(MessageContext synCtx) {
...
-                MessageContext itereatedMsgCtx =
+                MessageContext iteratedMsgCtx =
                         getIteratedMessage(synCtx, msgNumber++, msgCount, envelope, (OMNode) o);
                 ContinuationStackManager.
-                        addReliantContinuationState(itereatedMsgCtx, 0, getMediatorPosition());
-                target.mediate(itereatedMsgCtx);
+                        addReliantContinuationState(iteratedMsgCtx, 0, getMediatorPosition());
+
+                if(target.isAsynchronous()) {
+                    target.mediate(iteratedMsgCtx);
```
chanakaudaya: Don't we need to add try catch block to this section of the code. If something goes wrong here (SynapseException) it will not be captured.

virajsenevirathne: I added missing catch statements to the code. Please review.

virajsenevirathne:
Some formatting issues were found and fixed in https://github.com/wso2/wso2-synapse/pull/382

PR #285, opened 2015-07-26
Adding fix for ESBJAVA-2844- Had to set the SOAP action in message Co…

nadeeshaan:
Had to set the SOAP action in message Co…


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
```diff
@@ -380,6 +380,7 @@ public  void processEntityEnclosingRequest(MessageContext msgContext, boolean in
                 }
 
                 msgContext.setEnvelope(envelope);
+                msgContext.setSoapAction(soapAction);
```
chanakaudaya: Is there any chance that soapAction become null? If then, we need to check for null before setting the soapAction.

PR #295, opened 2015-07-30
https://wso2.org/jira/browse/ESBJAVA-3980

virajsenevirathne:
Fix File polling proxy creating too many ftp connections issue


modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -751,7 +751,14 @@ private void processFile(PollTableEntry entry, FileObject file) throws AxisFault
             		+ VFSUtils.maskURLPassword(file.toString()), e);
 
         } finally {
-           closeFileSystem(file);
```
malakasilva: You need to add this at root level

virajsenevirathne: It was closed at the root in a fix done by chanaka in ESB 4.9.0. Therefore fix only applies to here.

PR #296, opened 2015-07-31
fix for ESBJAVA-3997

bsenduran:
fix for [ESBJAVA-3997](https://wso2.org/jira/browse/ESBJAVA-3997): java.lang.ClassCastException when adding "Non retry http status codes" to a message processors


modules/core/src/main/java/org/apache/synapse/message/processor/impl/forwarder/ForwardingService.java
```diff
@@ -358,7 +358,7 @@ public void init(SynapseEnvironment se) {
         if (nonRetryStatusCodes != null) {
             for (String code : nonRetryStatusCodes) {
                 try {
-                    int codeI = Integer.parseInt(code);
+                    int codeI = Integer.parseInt(code.trim());
```
bsenduran: This fix is not the part of the addressed issue, but since there is a possibility to enter the comma separate values with spaces [ eg: 304, 305]  making this improvement

PR #309, opened 2015-08-13
ESBJAVA-4016 Exception is not shown when MP is deleted.

ravindraranwala:


modules/core/src/main/java/org/apache/synapse/startup/quartz/QuartzTaskManager.java
```diff
@@ -486,4 +486,8 @@ public boolean isTaskRunning(String taskName) {
     public void sendClusterMessage(Callable<Void> task) {
 
     }
+
+    public boolean isTaskExist(String arg0) {
+        return false;
```
chanakaudaya: @ravindraranwala 
What is the reason for returning false?

ravindraranwala: We haven't implemented QUartz related methods since we don't use them, instead we use Ntask components to get our work done. If some one need to use Quartz directly, they need to implement these functionalities using Quartz. Here we have just given a dummy implementation to get rid of compilation errors.

PR #313, opened 2015-08-15
Fix for the ESBJAVA-3852

prabathariyaratna:
Removed ManagedLifecycle implementation form the StoreMediator to avoid the artifact deployment sequnce issue


modules/core/src/main/java/org/apache/synapse/mediators/store/MessageStoreMediator.java
```diff
@@ -74,9 +67,18 @@ public boolean mediate(MessageContext synCtx) {
 
 
         if(synCtx != null) {
-
+            MessageStore messageStore = synCtx.getConfiguration().getMessageStore(messageStoreName);
             if(messageStore != null) {
 
+                if (messageStore.getParameters().get(PRODUCER_GUARANTEED_DELIVERY) != null) {
```
dinusha-dilrukshi: Better assign messageStore.getParameters().get(PRODUCER_GUARANTEED_DELIVERY) to a variable and reuse.

PR #322, opened 2015-08-22
Fix for the ESBJAVA-4091

prabathariyaratna:


prabathariyaratna:
Regression issue has reported for this changes and decided not to merge because we are no longer recommend to use legacy implementation

PR #337, opened 2015-09-01
Fix For ESBJAVA-4133

nadeeshaan:


modules/core/src/main/java/org/apache/synapse/deployers/AbstractSynapseArtifactDeployer.java
```diff
@@ -572,4 +572,8 @@ private String updateDefaultSequence(String filename, OMElement element, Propert
     public void setCustomLog (String artifactContainerName, String tenantId) {
         customLogContent = "[ Deployed From Artifact Container: " + artifactContainerName + " ] ";
     }
+
+    public void setCustomLogContentNull () {
+        customLogContent = null;
+    }
```
kasun04: Is this a proper method name?

nadeeshaan: Refined with [1] and [2]
[1] https://github.com/wso2/wso2-synapse/pull/400
[2] https://github.com/wso2/carbon-mediation/pull/482

PR #343, opened 2015-09-02
Esbjava 4130

virajsenevirathne:
Fix when multiple consumers are listening to the same location


virajsenevirathne:
Some formatting issues were found and fixed in https://github.com/wso2/wso2-synapse/pull/382

PR #379, opened 2015-10-20
Fix for URI Templates to seperate path params with ','

hevayo:


hevayo:
This fix will allow URI Template to be written as "/{param1},{param2},{param3}"  which will give same level path parameters.

PR #396, opened 2015-11-17
Fix for ESBJAVA-4305

kishokpriya:
If the argument value contain escape character, JSON payload generated by the mediator becomes malformed.


bsenduran:
fixed in https://github.com/wso2/wso2-synapse/pull/467

PR #409, opened 2015-12-04
ESBJAVA-4260: Allow optional query parameters to URI templates

hevayo:


balajisetty:
I was looking for this fix. Without going through entire code base, can any one brief on how to specify optional parameter? 

I tried uri,var.* and query.param.* both doesn't work

PR #411, opened 2015-12-14
send empty content to client application for 204 no content situation

laki88:


bsenduran:
can you please point to the public jira associated with the JIRA

laki88:
public jira is https://wso2.org/jira/browse/ESBJAVA-4386. this is resolved by the PR mentioned in jira.

PR #427, opened 2016-01-11
Improve the script mediator allow set properties with a scope

kishokpriya:


Buddhima:
Updated PR: https://github.com/wso2/wso2-synapse/pull/506

PR #495, opened 2016-03-25
Added an improvement to support DLC when we drop the message after ma…

shafreenAnfar:
…ximum number of retries in MP


shafreenAnfar:
There is a slight issue in the pull request. So closed it without merging.

PR #509, opened 2016-04-25
Fixing json when empty payload is sent with json content type

virajsenevirathne:
There are three main cases to be considered,

1. Different payload type is sent with JSON content type - With the fix it will throw an axis fault
2. Empty JSON payload sent with JSON content type & 
3. Non-empty JSON payload with JSON content type - Considered it as valid payload and continue with the processing


virajsenevirathne:
Related with https://github.com/wso2/wso2-synapse/pull/510

PR #519, opened 2016-05-03
Changes needed for mediation debugger wire logs

lankavitharana:


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/LoggingNHttpClientConnection.java
```diff
@@ -105,11 +108,83 @@ public void consumeInput(final NHttpClientEventHandler handler) {
         if (this.log.isDebugEnabled()) {
             this.log.debug(this.id + ": Consume input");
         }
-        super.consumeInput(handler);
-        if (isReleaseConn()) {
-            if (handler instanceof TargetHandler) {
-                ((TargetHandler) handler).getTargetConfiguration().getConnections().releaseConnection(this);
-                this.setReleaseConn(false);
+//        super.consumeInput(handler);
```
raviu: Why have we removed calling super.consumeInput()? Have we brought that into the logger class? Is it possible to do what we need only here and call super.consumeInput(), reason being, if there are fixes to HTTP Core library at this level, we may have to manually fix those in here. Is there a perf reason for doing this?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/LoggingNHttpClientConnection.java
```diff
@@ -119,7 +194,65 @@ public void produceOutput(final NHttpClientEventHandler handler) {
         if (this.log.isDebugEnabled()) {
             this.log.debug(this.id + ": Produce output");
         }
-        super.produceOutput(handler);
+//        super.produceOutput(handler);
```
raviu: Same comment as consumeInput() above.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/LoggingNHttpServerConnection.java
```diff
@@ -103,15 +107,152 @@ public void consumeInput(final NHttpServerEventHandler handler) {
         if (this.log.isDebugEnabled()) {
             this.log.debug(this.id + ": Consume input");
         }
-        super.consumeInput(handler);
+//        super.consumeInput(handler);
```
raviu: Same comment as for Client handler.

modules/core/src/main/java/org/apache/synapse/core/SynapseEnvironment.java
```diff
@@ -263,7 +263,7 @@
      *
      * @return whether debugging is enabled in the environment
      */
-    public boolean isDebugEnabled();
+    public boolean isDebuggerEnabled(); //todo refactor this "isDebuggerEnabled"
```
chanakaudaya: This todo item is no longer needed since it has already been done.

raviu:
This PR has been superseded by https://github.com/wso2/wso2-synapse/pull/530/

PR #530, opened 2016-05-06
Changes needed for mediation debugger wire logs

lankavitharana:


modules/core/src/main/java/org/apache/synapse/debug/constructs/SynapseMediationFlowPoint.java
```diff
@@ -64,6 +66,15 @@ public Mediator getMediatorReference() {
         return medRef;
     }
 
+//    public String getMediatorId() {
```
raviu: Shall we remove these comments?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/CalloutMediator.java
```diff
@@ -95,7 +95,8 @@
 
     public boolean mediate(MessageContext synCtx) {
 
-        if (synCtx.getEnvironment().isDebugEnabled()) {
+        if (synCtx.getEnvironment().isDebuggerEnabled()) {
+//            super.setWireLogHolderProperties(synCtx); //todo remove, this needs to be set only in mediators where outgoing messages are present
```
raviu: Remove Comment

PR #570, opened 2016-06-09
Fix for ESBJAVA-4664

isudana:


modules/core/src/main/java/org/apache/synapse/MessageContext.java
```diff
@@ -164,6 +164,15 @@
      */
     public Object getEntry(String key);
 
+
+    /**
+     * Get the value of a property set on the message instance or from the local registry
```
shafreenAnfar: Is it "local registry" or "local entry"

isudana: This is not the local registry space in wso2 products. 
In synapse, we call "local registry" which holds artifacts like local-entries, sequences, etc.

PR #584, opened 2016-06-23
Fixing the https://wso2.org/jira/browse/ESBJAVA-4704

chanakaudaya:
Original Fix was done by @isudana


dkgkim:
what maven version contains this changes? Is 'v2.1.7-wso2v3' version already released?
I think '2.1.7-wso2v3-BETA' version doesn't ..

PR #594, opened 2016-07-05
Removed setters that enabled timeoutOn and timeoutDuration to differ.

RawToast:
Previously, a developer would have to remember to set both variables when modifying this logic -- otherwise the application would not work as expected.


chanakaudaya:
@RawToast  Is there any public JIRA related to this fix?

RawToast:
@chanakaudaya Yes, I created the following Jira: https://wso2.org/jira/browse/APIMANAGER-5200

bsenduran:
@RawToast Thank you for your contribution

PR #668, opened 2016-11-02
Added the ability to configure enabled weak ciphers in ssl via axis2.xml

vinok88:
Preferred ciphers can be configured by adding the parameter <PreferredCiphers> with the comma separated names of preferred ciphers to the "https" transportSender or transportReceiver configs in axis2.xml or in inboundEndpoint configs.


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/util/NhttpConstants.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.synapse.transport.nhttp.util;
+
+/**
+ * This file holds all the constants related to NHTTP transports
+ */
+public class NhttpConstants {
```
isururanawaka: Why can't we use existing constant class ?

vinok88: Fixed. Moved to the existing class.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/ServerSSLSetupHandler.java
```diff
@@ -36,12 +36,16 @@
     /** Enabled SSL handshake protocols (e.g. SSLv3, TLSv1) */
     private final String[] httpsProtocols;
     private RevocationVerificationManager verificationManager;
+    /** Ciphers enabled in axis2.xml, enabled all if null*/
+    private final String[] preferredCiphers;
 
     public ServerSSLSetupHandler(final SSLClientAuth clientAuth, final String[] httpsProtocols,
-                                 final RevocationVerificationManager verificationManager) {
+                                 final RevocationVerificationManager verificationManager, final String[]
+            preferredCiphers) {
```
lankavitharana: formatting issue?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/config/ClientConnFactoryBuilder.java
```diff
@@ -154,6 +156,28 @@ public ClientConnFactoryBuilder parseSSL() throws AxisFault {
...
+        OMElement preferredCiphersEl = preferredCiphersParam != null ? preferredCiphersParam.getParameterElement() :
+                null;
+        String[] preferredCiphers = null;
+        final String configuredWeakCiphers =
+                preferredCiphersEl != null ? preferredCiphersEl.getText() : null;
+        if (configuredWeakCiphers != null && configuredWeakCiphers.trim().length() != 0) {
+            String[] configuredValues = configuredWeakCiphers.trim().split(",");
+            List<String> ciphersList = new ArrayList<String>(configuredValues.length);
+            for (String cipher : configuredValues) {
+                if (!cipher.trim().isEmpty()) {
```
akalankapagoda: We can trim the cipher first and assign it to itself so that we don't have to trim twice.

vinok88: Done. Thanks.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/config/ClientConnFactoryBuilder.java
```diff
@@ -154,6 +156,28 @@ public ClientConnFactoryBuilder parseSSL() throws AxisFault {
...
+                    ciphersList.add(cipher.trim());
+                }
+            }
+
+            preferredCiphers = ciphersList.toArray(new String[ciphersList.size()]);
+        }
+
+        if (preferredCiphers != null) {
```
akalankapagoda: We can remove the preferredCiphers variable and directly set SSL handler value in the place where preferredCiphers is initialized and remove this if block.

PR #669, opened 2016-11-03
ESBJAVA-4942 : Adding time out equal to socket time out for write condition await in…

nwnpallewela:
… the flush content method

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
@@ -21,6 +21,7 @@ import org.apache.http.nio.ContentDecoder;
 import org.apache.http.nio.ContentEncoder;
 import org.apache.http.nio.IOControl;
 import org.apache.http.params.HttpConnectionParams;
+import org.apache.http.params.HttpParams;
 import org.apache.synapse.transport.passthru.config.BaseConfiguration;
 import org.apache.synapse.transport.passthru.config.SourceConfiguration;
 import org.apache.synapse.transport.passthru.config.TargetConfiguration;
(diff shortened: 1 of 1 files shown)


vinok88:
Shall we mention the public Jira in the commit message.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
```diff
@@ -82,13 +103,32 @@ public Pipe(IOControl producerIoControl, ControlledByteBuffer buffer,
...
+
+    /**
+     * Extract socket time out value from baseConfig and return default value if not found
+     *
+     * @param baseConfig
+     * @return
+     */
+    private long getSocketTimeOut(BaseConfiguration baseConfig) {
+        if (baseConfig instanceof SourceConfiguration) {
+            return ((SourceConfiguration) baseConfig).getHttpParams()
```
lankavitharana: Better make sure "((SourceConfiguration) baseConfig).getHttpParams()" cannot be null when it comes to this.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
```diff
@@ -35,6 +40,8 @@
  */
 public class Pipe {
 
+    private static final int DEFAULT_TIME_OUT_VALUE = 180000;
```
isururanawaka: please add a comment here

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
```diff
@@ -82,13 +104,36 @@ public Pipe(IOControl producerIoControl, ControlledByteBuffer buffer,
...
+        this.socketTimeOut = getSocketTimeOut(baseConfig);
+    }
+
+    /**
+     * Extract socket time out value from baseConfig and return default value if not found
+     *
+     * @param baseConfig
+     * @return
+     */
+    private long getSocketTimeOut(BaseConfiguration baseConfig) {
```
isudana: Can't we use an existing utility for this ?

nwnpallewela: Used existing one and made the socket time out value static.
Thanks

PR #674, opened 2016-11-11
Fixing ESBJAVA-4891 & APIMANAGER-4200

madhawa-gunasekara:
This PR contains the fix to determine where we need to apply the consume and discard method. 
There are several incidents where we need to clear the buffer. like when the buffer is full. and also we need to clear the buffer if there is an error occurred during the middle of the message building.


madhawa-gunasekara:
JIRA links : https://wso2.org/jira/browse/ESBJAVA-4891
                  https://wso2.org/jira/browse/APIMANAGER-4200

madhawa-gunasekara:
@isudana & @isururanawaka Shall we merge this PR. Since this PR is needed for APIM release

PR #675, opened 2016-11-21
[ESBJAVA-3770] Dropping larger messages based on the user configuration

bsenduran:


bsenduran:
To enable message rejection add the following properties to the synapse.properties file 
* message.size.validation.enabled
* valid.max.message.size.in.bytes

Example of rejecting messages larger than 100kB 
```
valid.max.message.size.in.bytes=true
valid.max.message.size.in.bytes=100000
```

djkevincr:
This is already addressed by following PR https://github.com/wso2/wso2-synapse/pull/679  and hence closing the duplicate work.

djkevincr:
This is already addressed by following PR #679 and hence closing the duplicate work.

PR #679, opened 2016-12-01
ESBJAVA-3770 Message size based filtering to prevent OOM.

djkevincr:
Added support reject messages ( both request/response ) which exceeds a certain size threshold defined in passthru-http.properties

Eg:- 
message.size.validation.enabled=true
valid.max.message.size.in.bytes=100000

This threshold can deviate up to max = valid.max.message.size.in.bytes + io_buffer_size


modules/commons/src/main/java/org/apache/synapse/commons/util/MiscellaneousUtil.java
```diff
@@ -176,8 +176,14 @@ public static Properties loadProperties(String filePath) {
                 properties.load(in);
             } catch (IOException e) {
                 handleException("Error loading properties from a file at : " + filePath, e);
-            }
-        }
+            } finally {
+							  try {
```
pamod: Could you check the formatting

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/Pipe.java
```diff
@@ -242,6 +242,18 @@ public int produce(final ContentDecoder decoder) throws IOException {
         }
     }
 
+    public void forceProducerComplete(final ContentDecoder decoder) {
```
pamod: Could you include javaDocs ?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceHandler.java
```diff
@@ -83,6 +93,19 @@ public SourceHandler(SourceConfiguration sourceConfiguration) {
...
 
+        Properties props = MiscellaneousUtil.loadProperties(PROPERTY_FILE);
+        String validationProperty = MiscellaneousUtil.getProperty(props, MESSAGE_SIZE_VALIDATION, "false");
+        String validMaxMessageSizeStr = MiscellaneousUtil
+                .getProperty(props, VALID_MAX_MESSAGE_SIZE, String.valueOf(Integer.MAX_VALUE));
+        isMessageSizeValidationEnabled = Boolean.valueOf(validationProperty);
+        try {
+            validMaxMessageSize = Integer.valueOf(validMaxMessageSizeStr);
+        } catch (NumberFormatException e) {
+            log.warn("Invalid max message size configured for property \"valid.max.message.size.in.bytes\", "
```
pamod: is there any reason for not throwing the exception here ? also need to include the exception stacktrace ?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetHandler.java
```diff
@@ -390,6 +415,17 @@ private boolean handle202(MessageContext requestMsgContext) throws AxisFault {
         return false;
     }
 
+    private void dropTargetConnection(NHttpClientConnection conn) {
```
pamod: Good if we could include javaDocs describing the purpose of the method.

ramith: having a nice title to PR is good (as per guidelines set by git training done few weeks back)

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceHandler.java
djkevincr: +1, If a exception is thrown we assign a default value for validMaxMessageSize variable and continue the flow.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetHandler.java
djkevincr: +1

PR #687, opened 2017-01-16
[ESBJAVA-4989] Message size based filtering for nhttp transport

sdkottegoda:
Addresses the issue reported https://wso2.org/jira/browse/ESBJAVA-4989

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/ClientHandler.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/ClientHandler.java
@@ -158,6 +158,18 @@ public class ClientHandler implements NHttpClientEventHandler {
     private static final String CONTENT_TYPE = "Content-Type";
 
     /**
+     * Configuration which decides if message size validation is enabled/disabled. If enabled, messages of a size
+     * greater than 'validMaxMessageSize' will be dropped.
+     */
+    private static boolean isMessageSizeValidationEnabled = false;
+
+    /**
+     * The maximum messsage size that can be handled. This will only be valid, if 'isMessageSizeValidationEnabled' is
+     * enabled.
+     */
+    private static int validMaxMessageSize = Integer.MAX_VALUE;
+
+    /**
      * Create an instance of this client connection handler using the Axis2 configuration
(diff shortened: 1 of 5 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/ClientHandler.java
```diff
@@ -203,6 +215,16 @@ public ClientHandler(
 
         // set the latest openConnections map to MBean data during connection creation
         metrics.setConnectionsPerHosts(openConnections);
+
+        //load properties related to message size validation
+        isMessageSizeValidationEnabled = cfg.getMessageSizeValidationEnabled();
+        try {
```
vinok88: Reading the max message size is required only if isMessageSizeValidationEnabled is true. Shouldn't we put this try block inside a if(isMessageSizeValidationEnabled)?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/ClientHandler.java
```diff
@@ -158,6 +158,18 @@
     private static final String CONTENT_TYPE = "Content-Type";
 
     /**
+     * Configuration which decides if message size validation is enabled/disabled. If enabled, messages of a size
+     * greater than 'validMaxMessageSize' will be dropped.
+     */
+    private static boolean isMessageSizeValidationEnabled = false;
```
vinok88: Do we really need two properties here? IMO, validMaxMessageSize is sufficient. If it is not there, simply this feature is disabled. WDYT?

sdkottegoda: Yes, it is true. Thanks for pointing it out. But, we've set the same 2 properties in the passthru transport. Changing this will differ the behaviour of the same property in those 2 configs. WDYT? Should we change this? @djkevincr ?

vinok88: Agreed. I thought these are new properties.

PR #689, opened 2017-01-19
Fix problem with Service log not logging debug messages unless the de…

gstreetman:
…fault log is also set to debug.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/MediatorLog.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/MediatorLog.java
@@ -53,7 +53,14 @@ public class MediatorLog implements SynapseLog {
     }
 
     public boolean isDebugEnabled() {
-        return defaultLog.isDebugEnabled();
+    	// Returns true if either default or service log has debug enabled
+    	if (defaultLog.isDebugEnabled()) {
+    		return true;
+    	}
+    	if (synCtx.getServiceLog() != null && synCtx.getServiceLog().isDebugEnabled()) {
+    		return true;
+    	}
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/MediatorLog.java
```diff
@@ -114,8 +121,10 @@ public void auditLog(Object msg) {
      * Log a message at level DEBUG to all available/enabled logs.
      */
     public void auditDebug(Object msg) {
-        defaultLog.debug(msg);
-        if (synCtx.getServiceLog() != null) {
+    	if (defaultLog.isDebugEnabled()) {
```
vinok88: This if condition is not required, isDebugEnabled is checked inside the debug() method.

gstreetman: Actually it is needed. The  other change in this pull request is to change the MediatorLog isDebugEnabled function to return true if either defaultLog or the Service Log is enabled. When we get here, at least one of the two logs are enabled, but we do not know that both are enabled. So I used the isDebugEnabled function to keep from calling the debug function of the org.apache.commins.Log object, as is standard practice.

vinok88: Yes, you need this defaultLog.debug(msg) part. In standard practice, isDebugEnabled is used to avoid unnecessary string processing before calling actual debug() method. If there are no such string processing, having an if() clause is a waste.
Anyway, this is a just a minor concern :) I'll merge this PR!

PR #714, opened 2017-03-10
EI-#281:Fixing DevStudio CAPP deployment issue

erandacr:


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelaySecuirtyMessageBuilderDispatchandler.java
```diff
@@ -81,20 +87,25 @@ public InvocationResponse invoke(MessageContext messageContext) throws AxisFault
 					    serviceOpPart != null) {
 						axisService = registry.getService(serviceOpPart);
 						if (axisService != null) {
-							Parameter parameter = axisService.getParameter(SERVICE_TYPE);
-							if (parameter != null) {
-								if (!parameter.getValue().equals(PROXY)) {
+							isAdminService = ((axisService.getParameter(ADMIN_SERVICE) != null)
+									? Boolean.TRUE.toString().equals(axisService.getParameter(ADMIN_SERVICE).getValue())
```
vinok88: Don't we need to have equalsIgnoreCase ?

PR #721, opened 2017-03-24
ESBJAVA-5051 : MessageConverter Cannot create Message Context. Error:…

chanikag:
…null when scheduled message forwarding processor tries to fetch the message from message store and forward

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
@@ -48,8 +48,6 @@ import org.apache.synapse.commons.json.JsonUtil;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.util.UUIDGenerator;
 
-import javax.xml.stream.XMLStreamException;
-import javax.xml.stream.XMLStreamReader;
 import java.io.ByteArrayInputStream;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
```diff
@@ -131,6 +132,10 @@ public static MessageContext toMessageContext(StorableMessage message,
                 axis2Ctx.setOperationContext(operationContext);
                 axis2Ctx.setAxisService(axisService);
                 axis2Ctx.setAxisOperation(axisOperation);
+            } else {
+                if (logger.isDebugEnabled()) {
```
milindaperera: IMHO no need if check since no string concatenation in the debug log

chanikag: True. But as a practice, it is better to keep it in this way, IMO. Anyway this condition will be checked when the logger.debug() executes, even we remove the "if (logger.isDebugEnabled())". Therefore, it does not create any difference. What do you think?

milindaperera: agreed that make sense. 
I suggested that thinking to reduce context switching. when looking into the source of logger wrapping with if seems more efficient
rough analysis
with if (logger.isDebugEnabled())      :   if ->  (X * context switch) -> if 
without if (logger.isDebugEnabled()) :   context switch -> if -> (X * context switch) -> if

leave unchanged. 
sorry for noise

PR #726, opened 2017-03-31
Make escaping quotes of JSON configurable

VijithaEkanayake:
Fixes the issue where double quotes are escaped when a JSON double quoted string value is passed through arguments to payload mediator. Need to add below property to disable escaping quotes of valid JSON strings.
 ```xml
<property name="QUOTE_STRING_IN_PAYLOAD_FACTORY_JSON" value="true"/>
```
This is to make change done in [ESBJAVA-5053]  to be configurable as it breaks the functionality of some connectors.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
@@ -362,11 +362,15 @@ public class PayloadFactoryMediator extends AbstractMediator {
                     if (mediaType.equals(JSON_TYPE) && inferReplacementType(replacementEntry).equals(STRING_TYPE) &&
                             (!trimmedReplacementValue.startsWith("{") && !trimmedReplacementValue.startsWith("["))) {
                         replacementValue = escapeSpecialChars(replacementValue);
-                        // skip double quotes if replacement is boolean or null or valid json number
-                        if (!trimmedReplacementValue.equals("true") && !trimmedReplacementValue.equals("false")
-                            && !trimmedReplacementValue.equals("null") && !validJsonNumber
-                                .matcher(trimmedReplacementValue).matches()) {
-                            replacementValue = "\"" + replacementValue + "\"";
+                        // Check for following property which will force the string to include quotes
+                        Object force_string_quote = synCtx.getProperty("QUOTE_STRING_IN_PAYLOAD_FACTORY_JSON");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java [hunk 1]
```diff
@@ -362,11 +362,15 @@ private void replace(String format, StringBuffer result, MessageContext synCtx)
                     if (mediaType.equals(JSON_TYPE) && inferReplacementType(replacementEntry).equals(STRING_TYPE) &&
                             (!trimmedReplacementValue.startsWith("{") && !trimmedReplacementValue.startsWith("["))) {
                         replacementValue = escapeSpecialChars(replacementValue);
-                        // skip double quotes if replacement is boolean or null or valid json number
-                        if (!trimmedReplacementValue.equals("true") && !trimmedReplacementValue.equals("false")
-                            && !trimmedReplacementValue.equals("null") && !validJsonNumber
-                                .matcher(trimmedReplacementValue).matches()) {
-                            replacementValue = "\"" + replacementValue + "\"";
+                        // Check for following property which will force the string to include quotes
+                        Object force_string_quote = synCtx.getProperty("QUOTE_STRING_IN_PAYLOAD_FACTORY_JSON");
```
milindaperera: Better if "QUOTE_STRING_IN_PAYLOAD_FACTORY_JSON" defined as constant

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -362,11 +362,15 @@ private void replace(String format, StringBuffer result, MessageContext synCtx)
...
+                        if (force_string_quote != null && ((String) force_string_quote).equalsIgnoreCase("true")) {
+                            // skip double quotes if replacement is boolean or null or valid json number
+                            if (!trimmedReplacementValue.equals("true") && !trimmedReplacementValue.equals("false")
```
milindaperera: Possible to merge nested if statement to parent if

himasha: Merging might not be needed because the nested if does not need to be checked if the property is not available/value, is not true

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
erandacr: +1

PR #728, opened 2017-04-11
Fix ESBJAVA-5077

bsenduran:
Fix Prepared statement not closed properly, diff by @sajithshn

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/db/DBLookupMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/db/DBLookupMediator.java
@@ -41,6 +41,7 @@ public class DBLookupMediator extends AbstractDBMediator {
         // set as message context properties, any results that have been specified
         Connection con = null;
         ResultSet rs = null;
+        PreparedStatement ps = null;
         boolean threadInTx = false;
         try {
         	if(TranscationManger.isThreadHasEnlistment()){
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/db/DBLookupMediator.java
```diff
@@ -112,6 +113,11 @@ protected void processStatement(Statement stmnt, MessageContext msgCtx) {
                     rs.close();
                 } catch (SQLException e) {}
             }
+            if (ps != null) {
+                try {
+                    ps.close();
+                } catch (SQLException ignore) {}
```
madhawa-gunasekara: shall we print a warning message, without ignoring the exception

bsenduran: @madhawa-gunasekara Good point, but all the other similar close statements are ignoring the exception, hence ignored to have a consistence. IMO, if we added a warning we may need to add to the other close statements as well.

madhawa-gunasekara: yes, let's add to other close statements as well.

bsenduran: sure, we will keep track on this and add warning logs soon

PR #731, opened 2017-04-18
[APIMANAGER-4468] Throw RuntimeException instead of Error for invalid Json

manoramahp:
This fix is to change throwing RuntimeException instead of Error when received an invalid Json payload, as Synapse requires exception to be thrown
Related JIRA : https://wso2.org/jira/browse/APIMANAGER-4468

### DIFF
--- a/modules/commons/src/test/java/org/apache/synapse/commons/json/JsonStreamingBuilderTest.java
+++ b/modules/commons/src/test/java/org/apache/synapse/commons/json/JsonStreamingBuilderTest.java
@@ -19,8 +19,36 @@
 package org.apache.synapse.commons.json;
 
 import junit.framework.TestCase;
+import org.apache.axiom.om.OMElement;
+import org.apache.axis2.builder.Builder;
+import org.apache.axis2.context.MessageContext;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import java.io.InputStream;
 
 public class JsonStreamingBuilderTest extends TestCase {
     public void testCase() {
+        testInvalidJson(invalidJson);
+    }
+
+    private Builder jsonBuilder = Util.newJsonStreamBuilder();
+    private static final Log log = LogFactory.getLog(JsonStreamingBuilderTest.class);
+
+    private static final String invalidJson = "{\n" +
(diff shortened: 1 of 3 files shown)


modules/commons/src/test/java/org/apache/synapse/commons/json/JsonStreamingBuilderTest.java
```diff
@@ -19,8 +19,36 @@
 package org.apache.synapse.commons.json;
 
 import junit.framework.TestCase;
+import org.apache.axiom.om.OMElement;
+import org.apache.axis2.builder.Builder;
+import org.apache.axis2.context.MessageContext;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import java.io.InputStream;
```
milindaperera: Shall we add little description about the test case even though this is an old empty test-case

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonStreamingBuilderTest.java
```diff
@@ -19,8 +19,36 @@
...
+
+    public  void testInvalidJson(String jsonIn) {
+        try {
+            MessageContext message = Util.newMessageContext();
+            InputStream inputStream = Util.newInputStream(jsonIn.getBytes());
+            OMElement element  = jsonBuilder.processDocument(inputStream, "application/json", message);
+            message.getEnvelope().getBody().addChild(element);
+            log.info(message.getEnvelope().getBody().toString());
+        } catch (Exception e) {
+            assertEquals(e.getMessage(),"Error: could not match input");
```
milindaperera: are we testing the exception here? no asserts in try section

manoramahp: yes, we need to test whether we're getting an exception

milindaperera: Are we testing whether the runtime exception is thrown correctly? if so we can check instance of RuntimeException

manoramahp:
Opened https://github.com/wso2/wso2-synapse/pull/818 with the suggested changes

PR #734, opened 2017-04-20
Fix ClientWorker thread waiting issue.

djkevincr:
Related public JIRA : https://wso2.org/jira/browse/ESBJAVA-5071.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiver.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiver.java
@@ -78,7 +78,18 @@ public class ProxyServiceMessageReceiver extends SynapseMessageReceiver {
                 (mc.getWSAAction() != null ? mc.getWSAAction() : "null")));
 
             if (traceOn && trace.isTraceEnabled()) {
-                String[] cids = mc.getAttachmentMap().getAllContentIDs();
+                String[] cids = null;
+                try {
+                    cids = mc.getAttachmentMap().getAllContentIDs();
+                } catch (Exception ex){
+                    //partially read stream could lead to corrupted attachment map and hence this exception
+                    //corrupted attachment map leads to inconsistent runtime exceptions and behavior
+                    //discard the attachment map for the fault handler invocation
+                    //ensure the successful completion for fault handler flow
+                    mc.setAttachmentMap(null);
+                    log.debug("Synapse encountered an exception when reading attachments from bytes stream. " +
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiver.java
```diff
@@ -78,7 +78,18 @@ public void receive(org.apache.axis2.context.MessageContext mc) throws AxisFault
...
+                String[] cids = null;
+                try {
+                    cids = mc.getAttachmentMap().getAllContentIDs();
+                } catch (Exception ex){
+                    //partially read stream could lead to corrupted attachment map and hence this exception
+                    //corrupted attachment map leads to inconsistent runtime exceptions and behavior
+                    //discard the attachment map for the fault handler invocation
+                    //ensure the successful completion for fault handler flow
+                    mc.setAttachmentMap(null);
+                    log.debug("Synapse encountered an exception when reading attachments from bytes stream. " +
```
Asitha: Shall we add the exception as well to the log?

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
```diff
@@ -305,7 +309,18 @@ private void handleMessage(String messageID ,MessageContext response,
...
+                try {
+                    cids = response.getAttachmentMap().getAllContentIDs();
+                } catch (Exception ex){
+                    //partially read stream could lead to corrupted attachment map and hence this exception
+                    //corrupted attachment map leads to inconsistent runtime exceptions and behavior
+                    //discard the attachment map for the fault handler invocation
+                    //ensure the successful completion for fault handler flow
+                    response.setAttachmentMap(null);
+                    log.debug("Synapse encountered an exception when reading attachments from bytes stream. " +
+                            "Hence Attachments map is dropped from the message context.");
```
Asitha: Shall we log the exception as well? BTW shouldn't this be an error log?

modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiver.java
djkevincr: +1

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
djkevincr: +1

PR #746, opened 2017-05-23
[ESBJAVA-5098] Fix for SOAPHeaderBlock cast error with WS-Addressing

prabushi:


modules/core/src/main/java/org/apache/synapse/mediators/transform/HeaderMediator.java
```diff
@@ -230,15 +234,35 @@ private void addCustomHeader(MessageContext synCtx, String value) {
...
+            factory = OMAbstractFactory.getSOAP12Factory();
+        }
+        for (OMElement headerElement : headerElements) {
+            try {
+                newHeaderNodes.add(ElementHelper.toSOAPHeaderBlock(headerElement, factory));
+            } catch (Exception e) {
+                log.error("Unable to convert to SoapHeader Block", e);
+            }
+        }
+        for (OMNode newHeaderNode : newHeaderNodes) {
```
milindaperera: Why do we need two for loops? Shall we merge two for loops to one?

milindaperera: Also seems we can remove List<OMNode> newHeaderNodes also

prabushi: Moved to one loop.

PR #752, opened 2017-06-19
Fix the issue of changing the auth header order

sohaniwso2:
This fixes the issue of changing the order of the auth headers before sending the request to the backend

Public Jira  - https://wso2.org/jira/browse/ESBJAVA-5121

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
@@ -53,10 +53,10 @@ import java.io.OutputStream;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
 
 /**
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
```diff
@@ -124,8 +124,9 @@ public void start(NHttpClientConnection conn) throws IOException, HttpException
 
         long contentLength = -1;
         String contentLengthHeader = null;
-	    if(headers.get(HTTP.CONTENT_LEN) != null && headers.get(HTTP.CONTENT_LEN).size() > 0) {
-	        contentLengthHeader = headers.get(HTTP.CONTENT_LEN).first();
+	
+	    if(headers.get(HTTP.CONTENT_LEN) != null && headers.get(HTTP.CONTENT_LEN).iterator().hasNext()) {
```
madhawa-gunasekara: Shall we assign this value "headers.get(HTTP.CONTENT_LEN)" to variable and use

sohaniwso2: done at https://github.com/wso2/wso2-synapse/pull/753

PR #755, opened 2017-06-22
Fix local transport XA Transactions

madhawa-gunasekara:


modules/commons/src/main/java/org/apache/synapse/commons/transaction/TranscationManger.java
```diff
@@ -388,12 +388,7 @@ public static Transaction getTransaction() throws Exception {
 		    if (log.isDebugEnabled()) {
 			    log.debug("getTransaction Called");
 		    }
-
-		    TransactionManager txMgr = txManagers.get().get(key);
-		    txMgr.begin();
-		    Transaction tx = txMgr.getTransaction();
-		    transactions.get().put(key, tx);
-		    return tx;
+		    return transactions.get().get(key);
```
djkevincr: What if transactions.get().get(key) is null?

modules/commons/src/main/java/org/apache/synapse/commons/transaction/TranscationManger.java
```diff
@@ -388,12 +388,7 @@ public static Transaction getTransaction() throws Exception {
 		    if (log.isDebugEnabled()) {
 			    log.debug("getTransaction Called");
 		    }
-
-		    TransactionManager txMgr = txManagers.get().get(key);
-		    txMgr.begin();
```
djkevincr: Have you run all the tests? There may be invocations getTransaction() which requires txMgr.begin(). How do you handle such cases?

djkevincr:
I am approving this since, we got this reviewed and approved at support PR.

madhawa-gunasekara:
Resolves https://github.com/wso2/product-ei/issues/683

PR #756, opened 2017-06-27
Fix the Aggregate Mediator issue of not GC messages

sohaniwso2:
This fixes the issue of Messages not getting garbage collected when the messages do not match with the aggregate expression

Public Jira - https://wso2.org/jira/browse/ESBJAVA-5129


modules/core/src/main/java/org/apache/synapse/mediators/eip/aggregator/AggregateMediator.java
```diff
@@ -688,4 +688,13 @@ public void setComponentStatisticsId(ArtifactHolder holder) {
         }
         StatisticIdentityGenerator.reportingFlowContinuableEndEvent(mediatorId, ComponentType.MEDIATOR, holder);
     }
+    
+    private void handleException(Aggregate aggregate, String msg, Exception e, MessageContext msgContext) {
+        aggregate.clear();
+        if (e != null) {
```
milindaperera: single character variable name

sohaniwso2: fixed

PR #763, opened 2017-07-10
Add mediation security component to secure passwords for JMS Message Store

Nirthika:
ESBJAVA-2467 Use Secure vault tool to secure passwords in JMS Message Store instead of plain text

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
@@ -55,7 +55,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class JmsStore extends AbstractMessageStore {
-    protected static final Log log = LogFactory.getLog(InboundEndpoint.class);
+    protected static final Log log = LogFactory.getLog(JmsStore.class);
 
     /** JMS Broker username */
     public static final String USERNAME = "store.jms.username";
@@ -653,6 +653,11 @@ public class JmsStore extends AbstractMessageStore {
         return true;
     }
 
+    /**
+     * Use secure vault to secure password in JMS Message Store
+     * @param newParamValue
+     * @return
+     */
     private String resolveSecureVaultExpressions(String newParamValue) {
         Pattern vaultLookupPattern = Pattern.compile(secureVaultRegex);
         Matcher lookupMatcher = vaultLookupPattern.matcher(newParamValue);
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
@@ -26,10 +26,14 @@ import org.apache.synapse.config.SynapseConfiguration;
 import org.apache.synapse.core.SynapseEnvironment;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.core.axis2.Axis2SynapseEnvironment;
+import org.apache.synapse.inbound.InboundEndpoint;
+import org.apache.synapse.mediators.Value;
 import org.apache.synapse.message.MessageConsumer;
 import org.apache.synapse.message.MessageProducer;
 import org.apache.synapse.message.store.AbstractMessageStore;
 import org.apache.synapse.message.store.Constants;
+import org.apache.synapse.util.xpath.SynapseXPath;
+import org.jaxen.JaxenException;
 
 import javax.jms.Connection;
 import javax.jms.Destination;
@@ -47,8 +51,12 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class JmsStore extends AbstractMessageStore {
+    protected static final Log log = LogFactory.getLog(InboundEndpoint.class);
+
     /** JMS Broker username */
     public static final String USERNAME = "store.jms.username";
     /** JMS Broker password */
@@ -112,6 +120,10 @@ public class JmsStore extends AbstractMessageStore {
     private MessageProducer cachedProducer;
 
     private MessageProducer producer = null;
+
+    private SynapseEnvironment synapseEnvironment;
+    /** regex for secure vault expression */
+    private static final String secureVaultRegex = "\\{(wso2:vault-lookup\\('(.*?)'\\))\\}";
 
     public MessageProducer getProducer() {
         if (cacheLevel == 1 && cachedProducer != null) {
@@ -263,6 +275,7 @@ public class JmsStore extends AbstractMessageStore {
     } /** End of unsupported operations. */
 
     public void init(SynapseEnvironment se) {
+        synapseEnvironment = se;
         if (se == null) {
             logger.error("Cannot initialize store.");
             return;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -47,8 +51,12 @@
 import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class JmsStore extends AbstractMessageStore {
+    protected static final Log log = LogFactory.getLog(InboundEndpoint.class);
+
```
chanikag: Shouldn't this log is of this class (JmsStore)?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
         return true;
     }
 
+    /**
+     * Use secure vault to secure password in JMS Message Store
+     * @param newParamValue
+     * @return
```
abeykoon: It is good to mention what is returned here?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
...
+     */
+    private String resolveSecureVaultExpressions(String newParamValue) {
```
abeykoon: Let us put meaningful parameter name for "newParamValue"

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -113,6 +121,10 @@
 
     private MessageProducer producer = null;
 
+    private SynapseEnvironment synapseEnvironment;
+    /** regex for secure vault expression */
+    private static final String secureVaultRegex = "\\{(wso2:vault-lookup\\('(.*?)'\\))\\}";
```
abeykoon: Generally constants are denoted by capital letters. 
SECURE_VAULT_REGEX

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
...
+            //getting the expression with out curly brackets
+            String expressionStr = lookupMatcher.group(1);
+            try {
+                expression = new Value(new SynapseXPath(expressionStr));
+            } catch (JaxenException e) {
+                log.error("Error while building the expression : " + expressionStr);
+            }
+            if (expression != null) {
+                String resolvedValue = expression.evaluateValue(synapseEnvironment.createMessageContext());
+                if (resolvedValue == null || resolvedValue.isEmpty()) {
```
abeykoon: String.isEmprty() can be used here?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
...
+        if(lookupMatcher.find()) {
+            Value expression = null;
+            //getting the expression with out curly brackets
+            String expressionStr = lookupMatcher.group(1);
+            try {
+                expression = new Value(new SynapseXPath(expressionStr));
+            } catch (JaxenException e) {
+                log.error("Error while building the expression : " + expressionStr);
+            }
+            if (expression != null) {
```
abeykoon: Should we give out a warn message/error if expression is null?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -561,7 +574,8 @@ private boolean initme() {
             }
         }
         userName = (String) parameters.get(USERNAME);
-        password = (String) parameters.get(PASSWORD);
+        password = resolveSecureVaultExpressions((String) parameters.get(PASSWORD));
```
abeykoon: What if password parameter is not encrypted? Can resolveSecureVaultExpressions method handle that too?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
...
+        Matcher lookupMatcher = vaultLookupPattern.matcher(newParamValue);
+        //setting value initially
+        if(lookupMatcher.find()) {
+            Value expression = null;
+            //getting the expression with out curly brackets
+            String expressionStr = lookupMatcher.group(1);
+            try {
+                expression = new Value(new SynapseXPath(expressionStr));
+            } catch (JaxenException e) {
+                log.error("Error while building the expression : " + expressionStr);
```
abeykoon: It is better to print the stack trace as well?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
         return true;
     }
 
+    /**
+     * Use secure vault to secure password in JMS Message Store
+     * @param newParamValue
```
malakaganga: Can we add a description to parameter here ?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -639,6 +653,37 @@ private boolean initme() {
...
+     * @return
+     */
+    private String resolveSecureVaultExpressions(String newParamValue) {
+        Pattern vaultLookupPattern = Pattern.compile(secureVaultRegex);
+        Matcher lookupMatcher = vaultLookupPattern.matcher(newParamValue);
+        //setting value initially
+        if(lookupMatcher.find()) {
```
abeykoon: What if lookupMatcher.find() == false? Should code handle that, or inform that in the log/UI?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
afkham: Yes, this has to be changed since it is a coding style violation

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
afkham: If the code block below line 674 is moved to line 671, this null check won't be needed. Also if JaxenException gets thrown while building this xpath, it looks like an unrecoverable situation, so we should throw a runtime exception (may be SynapseException, which is a RuntimeException) and then init() will fail in line 577, which is the expected behavior.

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
Nirthika: There is no problem, if password parameter is not encrypted. resolveSecureVaultExpressions method can handle it by return the actual password.

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
isudana: null check is not needed if we throw a synapse exception within the catch block. I think its better not to move the code after line 674 into the try block.

PR #768, opened 2017-07-14
[ESBJAVA-5202] - Log mediator does not behave correctly when XML element is an element that stores a sourced JSON payload

Subasinghe:
This PR fixes the behavior of Log mediator for XML elements which store sourced JSON payload.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
@@ -197,7 +197,7 @@ public class LogMediator extends AbstractMediator {
         sb.append(getSimpleLogMessage(synCtx));
         try {
             org.apache.axis2.context.MessageContext a2mc = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-            if (JsonUtil.hasAJsonPayload(a2mc)) {
+            if (JsonUtil.hasAJsonPayload(a2mc) && synCtx.getEnvelope() == null) {
                 sb.append(separator).append("Payload: ").append(JsonUtil.jsonPayloadToString(a2mc));
             } else if (synCtx.getEnvelope() != null) {
                 sb.append(separator).append("Envelope: ").append(synCtx.getEnvelope());
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
@@ -197,6 +197,8 @@ public class LogMediator extends AbstractMediator {
         sb.append(getSimpleLogMessage(synCtx));
         try {
             org.apache.axis2.context.MessageContext a2mc = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
+            // If there is a XML element which store sourced JSON payload, synCtx.getEnvelope() will not be null.
+            // So need to check it also to avoid log in JSON format i.e. {"value" : "test"}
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -197,7 +197,7 @@ private String getFullLogMessage(MessageContext synCtx) {
         sb.append(getSimpleLogMessage(synCtx));
         try {
             org.apache.axis2.context.MessageContext a2mc = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-            if (JsonUtil.hasAJsonPayload(a2mc)) {
+            if (JsonUtil.hasAJsonPayload(a2mc) && synCtx.getEnvelope() == null) {
```
abeykoon: Should we explain the reason for null check here?

a5anka:
It is better if we can add a small description to the commit message on what and why we did to fix the issue.

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
a5anka: Isn't it better to move the condition in else if to top? That way we don't have to add the extra check (synCtx.getEnvelope() == null)).

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -197,7 +197,9 @@ private String getFullLogMessage(MessageContext synCtx) {
         sb.append(getSimpleLogMessage(synCtx));
         try {
             org.apache.axis2.context.MessageContext a2mc = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-            if (JsonUtil.hasAJsonPayload(a2mc)) {
+            // If there is a XML element which store sourced JSON payload, synCtx.getEnvelope() will not be null.
+            // So need to check it also to avoid log in JSON format i.e. {"value" : "test"}
+            if (JsonUtil.hasAJsonPayload(a2mc) && synCtx.getEnvelope() == null) {
```
a5anka: Isn't it better to move the condition in else if to top? That way we don't have to add the extra check (synCtx.getEnvelope() == null)).

Subasinghe: Agreed

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -197,11 +197,14 @@ private String getFullLogMessage(MessageContext synCtx) {
         sb.append(getSimpleLogMessage(synCtx));
         try {
             org.apache.axis2.context.MessageContext a2mc = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-            if (JsonUtil.hasAJsonPayload(a2mc)) {
-                sb.append(separator).append("Payload: ").append(JsonUtil.jsonPayloadToString(a2mc));
-            } else if (synCtx.getEnvelope() != null) {
+            // If there is a XML element which store sourced JSON payload, synCtx.getEnvelope() will not be null.
+            // So need to check it also to avoid log in JSON format i.e. {"value" : "test"}
+            if (synCtx.getEnvelope() != null) {
                 sb.append(separator).append("Envelope: ").append(synCtx.getEnvelope());
```
isudana: This fix is not correct. This will not print JSON, even it is a real native JSON payload as synCtx.getEnvelope() will never become null.

PR #772, opened 2017-07-17
[ESBJAVA-4721] Implement a way to add PIs in JSON -> XML transformation

vinok88:
This introduces a new property
synapse.json.to.xml.processing.instruction.enabled to enable PI in JSON
-> XML transformation

### DIFF
--- a/modules/commons/src/main/java/org/apache/synapse/commons/json/JsonDataSource.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/json/JsonDataSource.java
@@ -151,6 +151,6 @@ final class JsonDataSource implements OMDataSource {
             throw new XMLStreamException("Could not reuse JSON stream from JSON Data Source.", e);
         }
         // Inform reader to add or remove PI's from XML output as per the configurations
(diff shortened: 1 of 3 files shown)


modules/commons/src/main/java/org/apache/synapse/commons/json/JsonDataSource.java
```diff
@@ -150,6 +150,7 @@ public XMLStreamReader getReader() throws XMLStreamException {
             logger.error("#getReader. Could not reuse JSON stream from JSON Data Source. Error>> " + e.getLocalizedMessage());
             throw new XMLStreamException("Could not reuse JSON stream from JSON Data Source.", e);
         }
-        return JsonUtil.getReader(inputStream, false); // Do not add PIs to the XML output of this reader
+        // Inform reader to add or remove PI's from XML output as per the configurations
+        return JsonUtil.getReader(inputStream, JsonUtil.isPiEnabled()); // Do not add PIs to the XML output of this reader
```
VijithaEkanayake: Shall We remove following comment (// Do not add PIs to the XML output of this reader)

PR #774, opened 2017-07-18
[ESBJAVA-5044] Error while connecting to samba server - A duplicate name exists on the network

milindaperera:
The hostname resolution for SMB file is failed due to a limitation in the jcifs library. hence need to resolve the hostname of SMB URI in Synapse level. Hostname resolution is done at deployment to improve performance and if the user needs to dynamically resolve the hostname (in case required to adapt with /etc/hosts changes), we will introduce new property "**transport.vfs.ResolveHostsDynamically**"

Property "transport.vfs.ResolveHostsDynamically" (Boolean):
- **true**: resolve hostname at runtime (Only effective for SMB protocol)
- **false** (Default): resolve hostname at deployment

Fix Jira URL: https://wso2.org/jira/browse/ESBJAVA-5044
Note: Merge https://github.com/wso2/product-ei/pull/769 befor merging this

### DIFF
--- a/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
@@ -485,10 +485,10 @@ public class VFSUtils {
 
     /**
      * Function to resolve hostname of the vfs uri
-     * @param uri
-     * @return
-     * @throws FileSystemException
-     * @throws UnknownHostException
+     * @param uri URI need to resolve
+     * @return hostname resolved uri
+     * @throws FileSystemException Unable to decode due to malformed URI
+     * @throws UnknownHostException Error occurred while resolving hostname of URI
      */
     public static String resolveUriHost (String uri) throws FileSystemException, UnknownHostException {
         return resolveUriHost(uri, new StringBuilder());
@@ -499,10 +499,10 @@ public class VFSUtils {
      * same uri provided for {uri}
      * Protocols resolved : SMB
      * @param uri URI need to resolve
-     * @param strBuilder
+     * @param strBuilder string builder to use to build the resulting uri
      * @return hostname resolved uri
-     * @throws FileSystemException
-     * @throws UnknownHostException
+     * @throws FileSystemException Unable to decode due to malformed URI
+     * @throws UnknownHostException Error occurred while resolving hostname of URI
      */
     public static String resolveUriHost (String uri, StringBuilder strBuilder)
             throws FileSystemException, UnknownHostException {
@@ -558,6 +558,7 @@ public class VFSUtils {
         int pos = 0;
         for (; pos < maxlen; pos++) {
             char ch = name.charAt(pos);
+            //if /;?:@&=+$, characters found means, we have passed the hostname, hence break
             if (ch == '/' || ch == ';' || ch == '?' || ch == ':'
                     || ch == '@' || ch == '&' || ch == '=' || ch == '+'
                     || ch == '$' || ch == ',') {
@@ -565,6 +566,7 @@ public class VFSUtils {
             }
         }
         if (pos == 0) {
+            //haven't found the hostname
             return null;
         }
 
--- a/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
@@ -483,6 +483,121 @@ public class VFSUtils {
         return opts;
     }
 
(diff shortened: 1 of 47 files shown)


modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
```diff
@@ -483,6 +483,121 @@ public static FileSystemOptions attachFileSystemOptions(Map<String, String> opti
         return opts;
     }
 
+    /**
+     * Function to resolve hostname of the vfs uri
+     * @param uri
+     * @return
+     * @throws FileSystemException
+     * @throws UnknownHostException
+     */
```
abeykoon: Shall we complete the comment?

modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
```diff
@@ -483,6 +483,121 @@ public static FileSystemOptions attachFileSystemOptions(Map<String, String> opti
...
+     * Extracts the hostname from a URI.  The scheme://userinfo@ part has
+     * been removed.
+     * extracted hostname will be reoved from the StringBuilder
+     */
+    private static String extractHostName(StringBuilder name) {
+        final int maxlen = name.length();
+        int pos = 0;
+        for (; pos < maxlen; pos++) {
+            char ch = name.charAt(pos);
+            if (ch == '/' || ch == ';' || ch == '?' || ch == ':'
```
abeykoon: Should we describe what we do here?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
```diff
@@ -197,43 +236,59 @@ public Integer getFileProcessingCount() {
...
+            this.moveAfterProcess = resolveHostAtDeployment(moveAfterProcess);
         }
     }
 
     public String getMoveAfterErrors() {
+        if (resolveHostsDynamically) {
+            try {
+                return VFSUtils.resolveUriHost(moveAfterErrors);
+            } catch (UnknownHostException | FileSystemException e) {
+                log.warn("Unable to resolving hostname of transport.vfs.MoveAfterErrors: " +
```
abeykoon: Is this a Warning or an actual error?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
```diff
@@ -138,10 +146,25 @@ public PollTableEntry(boolean fileLocking) {
     }
 
     public String getFileURI() {
-        return fileURI;
+        if (resolveHostsDynamically) {
+            try {
+                return VFSUtils.resolveUriHost(fileURI);
+            } catch (UnknownHostException | FileSystemException e) {
+                log.warn("Unable to resolving hostname of transport.vfs.FileURI : " + VFSUtils.maskURLPassword(fileURI), e);
```
vinok88: Shall we change this to "Unable to resolve the hostname of the ...."

modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
milindaperera: Thanks, will update

modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
milindaperera: Thanks for pointing, I stole the logic from wso2-commons-vfs :-D 
https://github.com/wso2/wso2-commons-vfs/blob/master/core/src/main/java/org/apache/commons/vfs2/provider/HostFileNameParser.java#L194

I'll describe the logic in comment as you suggest

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
milindaperera: Thanks, will update the message

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
milindaperera: According to pattern used in this module, warning is used in most similar cases. Here warn is used due to this exception doesn't halt the continiuation of the functionality and just to find root causse if failed later.

PR #780, opened 2017-08-02

Nirthika:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
@@ -21,7 +21,10 @@ package org.apache.synapse.config.xml;
 import org.apache.axiom.om.OMAbstractFactory;
 import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.Mediator;
+import org.apache.synapse.SynapseException;
 import org.jaxen.JaxenException;
 
 import javax.xml.namespace.QName;
@@ -34,6 +37,8 @@ import java.util.Properties;
 
 
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
+    private static final Log logger = LogFactory.getLog(EnrichMediatorFactory.class.getName());
+
     private static final QName XML_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "enrich");
     private static final QName ATT_PROPERTY = new QName("property");
     private static final QName ATT_XPATH = new QName("xpath");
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +194,35 @@ public QName getTagQName() {
...
+                logger.info("Un-expected target type");
+            }
+            if (targetType == 1) {
+                throw new SynapseException("Envelope not support for target attribute");
+            }
+        }
+
+        // check combination
+        if ((sourceType == 1 && targetType == 0) || (sourceType == 1 && targetType == 2) || (sourceType == 2
+                && targetType == 2)) {
```
Manuri: Can we make the comment more descriptive explaining what this condition really means?

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +194,35 @@ public QName getTagQName() {
...
+            if (sourceType < 0) {
+                logger.info("Un-expected source type");
+            }
+        }
+
+        // target type attribute
+        OMAttribute targetTypeAttr = targetEle.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
+            targetType = convertTypeToInit(targetTypeAttr.getAttributeValue());
+            if (targetType < 0) {
```
a5anka: It is better if we do not use magic numbers in conditions to improve readability. Maybe we can define a constant with value 0.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +194,35 @@ public QName getTagQName() {
...
+            sourceType = convertTypeToInit(sourceTypeAttr.getAttributeValue());
+            if (sourceType < 0) {
+                logger.info("Un-expected source type");
+            }
+        }
+
+        // target type attribute
+        OMAttribute targetTypeAttr = targetEle.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
+            targetType = convertTypeToInit(targetTypeAttr.getAttributeValue());
```
a5anka: Shouldn't this method name be `convertTypeToInt`?

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
a5anka: It is better if we do not use magic numbers in conditions to improve readability. Maybe we can define constants for these values.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: It was the existing implementation, so need to change in many classes. Clear comments are added to explaining what this condition really means.

PR #785, opened 2017-08-09
Add unit tests for nashorn java script engine

malakaganga:
Remove tests from test suite since the tests in
test suite repeat with each test class in module extension
and added them to a separate class.


modules/extensions/src/test/java/org/apache/synapse/mediators/bsf/javascript/NashornJavaScriptMediatorTest.java
```diff
@@ -0,0 +1,149 @@
...
+import javax.activation.DataSource;
+import java.util.LinkedHashMap;
+
+/**
+ * Test functions of Nashorn Java Script Mediator
+ */
+public class NashornJavaScriptMediatorTest extends TestCase {
+
+    private static final String inlinescript = "var state=5;";
+
```
chanikag: Better to define constants in Block Capitals

PR #789, opened 2017-08-10
[EI-TOOL-74] Fix for Log Separator in Log mediator

Bhashinee:
When you add "\t" in the log separator field in design view of the log mediator, it is not converted into relevant encoded characters when you switch into source view (Dev Studio). Fixed this issue.
Related issue : https://github.com/wso2/devstudio-tooling-ei/issues/74

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
@@ -244,13 +244,9 @@ public class LogMediator extends AbstractMediator {
     }
 
     public void setSeparator(String separator) {
-        if (separator.equals("\\n")) {
-            this.separator = "\n";
-        } else {
-            this.separator = separator;
-        }
+        this.separator = separator.replace("\\n", "\n").replace("\\t", "\t");
     }
-        
+
     public void addProperty(MediatorProperty p) {
         properties.add(p);
     }


modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -244,13 +244,9 @@ public String getSeparator() {
     }
 
     public void setSeparator(String separator) {
-        if (separator.equals("\\n")) {
-            this.separator = "\n";
-        } else {
-            this.separator = separator;
-        }
+        this.separator = separator.replace("\\n", "\n").replace("\\t", "\t");
```
abeykoon: What about other characters other than \n and \t? Is there any other possible characters? If there are, and next time when a user use such, we need to update the code again.

bsenduran: Yes, lets check the possible special characters and update this. 
I guess the best option would be providing a drop down to select the log separator, rather than providing a text box to enter it. But may be this is too late for this change

madhawa-gunasekara: IMO, This is should be handled in front end side. not in here. Can't we handle this in front end side?

madhawa-gunasekara:
Merged.

madhawa-gunasekara:
@Bhashinee Please move hardcoded strings into static variables in another pull request.

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
Bhashinee: We cannot handle this in front end.

PR #791, opened 2017-08-14
Fix preserving message processor state

bsenduran:
By introducing a state in registry, the message processor's state is perserved.
Issue: wso2/product-ei#849

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
@@ -20,6 +20,7 @@ package org.apache.synapse.message.processor.impl;
 
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Properties;
 import java.util.concurrent.Callable;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -81,6 +83,12 @@
 
 	private int memberCount = 1;
 
+	private Registry registry;
+
+	private static final String REG_PROCESSOR_BASE_PATH = "/repository/components/org.apache.synapse.message.processor/";
+
+	private static final String MP_STATE = "MESSAGE_PROCESSOR_STATE";
```
milindaperera: Shall we add little description about this property here?

bsenduran: isn't this property self descriptive enough

milindaperera: 👍

milindaperera: Any way, I felt better if we have at least what are the processor states without looking ussage of this property. That's why I suggested.

PR #795, opened 2017-08-15
Fix error when deleting a non existing resource

bsenduran:
Resolves https://github.com/wso2/product-ei/issues/872
by checking the resource existence before tyring to delete the resource

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
@@ -340,9 +340,9 @@ public abstract class ScheduledMessageProcessor extends AbstractMessageProcessor
             logger.info("Successfully destroyed message processor [" + getName() + "].");
         }
 
-		if(!preserveState) {
-			registry.delete(REG_PROCESSOR_BASE_PATH + getName());
-		}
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -549,4 +549,10 @@ private void setMessageProcessorState(ProcessorState state) {
 		registry.newNonEmptyResource(REG_PROCESSOR_BASE_PATH + getName(), false, "text/plain", state.toString(),
 				MP_STATE);
 	}
+
+    private void deleteMessageProcessorState() {
+        if (registry.getResourceProperties(REG_PROCESSOR_BASE_PATH + getName()) != null) {
+            registry.delete(REG_PROCESSOR_BASE_PATH + getName());
```
sdkottegoda: We have the same check inside WSO2Registry.delete() as well. Canwe remove that since it's handled from a higher level?

bsenduran: Unfortunately the WSO2Registry and ESBRegistry implementations are different on how the non existing resource is being deleted. Also the  WSO2Registry.delete() call be called from different  other components

PR #797, opened 2017-08-16
[EI-722] Fix expressions with nested brackets cannot be used with xPath 2.0

erandacr:
Expressions with nested brackets not used in XPath 1, but XPath 2.0 does. Since we are trying to compile every expression with jaxen first it will give an error as it doesn't support nested brackets.

Fixes https://github.com/wso2/product-ei/issues/722

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
@@ -23,7 +23,9 @@ import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
+import org.apache.synapse.config.SynapsePropertiesLoader;
 import org.apache.synapse.util.xpath.SynapseJsonPath;
 import org.apache.synapse.util.xpath.SynapseXPath;
 import org.jaxen.JaxenException;
@@ -48,7 +50,22 @@ public class SynapsePathFactory {
             if(pathAttrib.getAttributeValue().startsWith("json-eval(")) {
                 path = new SynapseJsonPath(pathAttrib.getAttributeValue().substring(10, pathAttrib.getAttributeValue().length() - 1));
             } else {
-                path = new SynapseXPath(pathAttrib.getAttributeValue());
+                try {
+                    path = new SynapseXPath(pathAttrib.getAttributeValue());
+                } catch (org.jaxen.XPathSyntaxException ex) {
+                    /* Try and see whether the expression can be compiled with XPath 2.0
+                     * This will only be done if the failover DOM XPath 2.0 is enabled */
+                    if (Boolean.parseBoolean(SynapsePropertiesLoader.loadSynapseProperties().
+                            getProperty(SynapseConstants.FAIL_OVER_DOM_XPATH_PROCESSING))) {
+                        if(log.isDebugEnabled()) {
+                            log.debug(
+                                    "Trying to compile the expression in XPath 2.0: " + pathAttrib.getAttributeValue());
+                        }
+                        path = new SynapseXPath(pathAttrib.getAttributeValue(), elem);
+                    } else {
+                        throw ex;
+                    }
+                }
             }
 
             OMElementUtils.addNameSpaces(path, elem, log);
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXPathFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXPathFactory.java
@@ -23,7 +23,9 @@ import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
+import org.apache.synapse.config.SynapsePropertiesLoader;
 import org.apache.synapse.util.xpath.SynapseXPath;
 import org.jaxen.JaxenException;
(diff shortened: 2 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
```diff
@@ -184,6 +198,53 @@ public SynapseXPath(String xpathString) throws JaxenException {
...
+        /* Handle special cases of content aware/unaware scenarios */
+        if (xpathExpr.contains("/")) {
+            contentAware = true;
+        } else if (xpathExpr.contains("get-property('From'") || xpathExpr.contains("get-property('FAULT')")) {
+            contentAware = true;
+        } else {
+            contentAware = false;
+        }
+
+        if(xpathExpr.contains("$trp") || xpathExpr.contains("$ctx") || xpathExpr.contains("$axis2")){
```
madhawa-gunasekara: Fix formatting

modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
```diff
@@ -48,7 +50,22 @@ public static SynapsePath getSynapsePath(OMElement elem, QName attribName)
             if(pathAttrib.getAttributeValue().startsWith("json-eval(")) {
                 path = new SynapseJsonPath(pathAttrib.getAttributeValue().substring(10, pathAttrib.getAttributeValue().length() - 1));
             } else {
-                path = new SynapseXPath(pathAttrib.getAttributeValue());
+                try {
+                    path = new SynapseXPath(pathAttrib.getAttributeValue());
+                } catch (org.jaxen.XPathSyntaxException ex) {
+                    /* Try and see whether the expression can be compiled with XPath 2.0
+                     * This will only be done if the failover DOM XPath 2.0 is enabled */
+                    if (Boolean.TRUE.toString().equalsIgnoreCase(SynapsePropertiesLoader.loadSynapseProperties().
```
bsenduran: ` Boolean.parseBoolean(..)` would be nice here

modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
```diff
@@ -48,7 +50,22 @@ public static SynapsePath getSynapsePath(OMElement elem, QName attribName)
...
+                            getProperty(SynapseConstants.FAIL_OVER_DOM_XPATH_PROCESSING))) {
+                        if(log.isDebugEnabled()) {
+                            log.debug(
+                                    "Trying to compile the expression in XPath 2.0: " + pathAttrib.getAttributeValue());
+                        }
+                        path = new SynapseXPath(pathAttrib.getAttributeValue(), elem);
+                    } else {
+                        throw ex;
```
abeykoon: Should we say, try enabling SynapseConstants.FAIL_OVER_DOM_XPATH_PROCESSING?

abeykoon: Make sure this should go to documentation.

modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXPathFactory.java
```diff
@@ -43,8 +45,21 @@ public static SynapseXPath getSynapseXPath(OMElement elem, QName attribName)
...
 
         if (xpathAttrib != null && xpathAttrib.getAttributeValue() != null) {
-
-            xpath = new SynapseXPath(xpathAttrib.getAttributeValue());
+            try {
+                xpath = new SynapseXPath(xpathAttrib.getAttributeValue());
+            } catch (org.jaxen.XPathSyntaxException ex) {
+                    /* Try and see whether the expression can be compiled with XPath 2.0
+                     * This will only be done if the failover DOM XPath 2.0 is enabled */
+                if (Boolean.TRUE.toString().equalsIgnoreCase(SynapsePropertiesLoader.loadSynapseProperties().
```
bsenduran: same as above

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
```diff
@@ -184,6 +198,53 @@ public SynapseXPath(String xpathString) throws JaxenException {
...
+        /* Make evaluation of this expression forced to XPath 2.0 */
+        this.forceFailoverEvaluation = Boolean.TRUE;
+
+        PassThroughConfiguration conf = PassThroughConfiguration.getInstance();
+        bufferSizeSupport =conf.getIOBufferSize();
+
+        /* Handle special cases of content aware/unaware scenarios */
+        if (xpathExpr.contains("/")) {
+            contentAware = true;
+        } else if (xpathExpr.contains("get-property('From'") || xpathExpr.contains("get-property('FAULT')")) {
```
milindaperera: we can merge if and else if statements to one since we only set contentAware = true;

modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXPathFactory.java
erandacr: Fixed

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
erandacr: fixed, thanks

modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
erandacr: fixed, thanks

modules/core/src/main/java/org/apache/synapse/config/xml/SynapsePathFactory.java
erandacr: this can happen even for the invalid XPath expressions. Should we log and throw the exception? And yeah, I will add this to the documents.
Thanks :)

PR #800, opened 2017-08-17
Fix trace log not printing in log mediator

bsenduran:
Fix https://wso2.org/jira/browse/ESBJAVA-5207
Fix the issue in trace logs are not getting printed when we use category=trace in log mediator
This can be easily observed in per API logs


milindaperera:
Better if we have little Description in PR

PR #804, opened 2017-08-21
Change getPayloadXml method to return OMElement

malakaganga:
Convert the return type of getPayloadXml method into OMElement.

Fix https://github.com/wso2/product-ei/issues/895

### DIFF
--- a/modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/NashornJavaScriptMessageContext.java
+++ b/modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/NashornJavaScriptMessageContext.java
@@ -97,14 +97,9 @@ public class NashornJavaScriptMessageContext implements ScriptMessageContext {
      * The payload is the first element inside the SOAP <Body> tags
      *
      * @return the XML SOAP Body
-     * @throws ScriptException in-case of an error in getting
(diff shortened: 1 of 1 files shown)


modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/NashornJavaScriptMessageContext.java
```diff
@@ -101,10 +101,7 @@ public NashornJavaScriptMessageContext(MessageContext mc, XMLHelper xmlHelper) {
      * the XML representation of SOAP Body payload
      */
     public Object getPayloadXML() throws ScriptException {
-        SOAPEnvelope envelope = mc.getEnvelope();
-        SOAPBody soapBody = envelope.getBody();
-        OMElement omElement = soapBody.getFirstElement();
-        return xmlHelper.toScriptXML(omElement);
+        return mc.getEnvelope().getBody().getFirstElement();
```
sdkottegoda: Can this method still throw a ScriptException?

a5anka: Do we need the ScriptException now?

PR #809, opened 2017-08-29
[EI-543] Fix incorrect printing of debug log

pamod:


modules/securevault/src/main/java/org/apache/synapse/securevault/secret/SecretManager.java
```diff
@@ -34,6 +34,11 @@
     private final static String PROP_PROVIDER = "provider";
     /* Dot string */
     private final static String DOT = ".";
+    /**
+     * Will represent empty string
+     */
+    private static final String EMPTY = "";
```
a5anka: Can we replace this with org.apache.commons.lang.StringUtils.EMPTY?

PR #812, opened 2017-09-04
Fix ESBJAVA-5216. Per-API Logs not generates after Call Mediator invoke.

wggihan:
Per-API logs not getting generate after the Call mediator invoke. 

Public JIRA: https://wso2.org/jira/browse/ESBJAVA-5216

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2MessageContext.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2MessageContext.java
@@ -41,6 +41,7 @@ import org.apache.synapse.core.SynapseEnvironment;
 import org.apache.synapse.endpoints.Endpoint;
 import org.apache.synapse.mediators.base.SequenceMediator;
 import org.apache.synapse.mediators.template.TemplateMediator;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2MessageContext.java
```diff
@@ -511,17 +512,22 @@ public Log getServiceLog() {
         if (serviceLog != null) {
             return serviceLog;
         } else {
+            String serviceLoggerName;
             String serviceName = (String) getProperty(SynapseConstants.PROXY_SERVICE);
+            // Added REST API service logger resolver to fix ESBJAVA-5216.
+            if (serviceName == null) {
+                serviceName = (String) getProperty(RESTConstants.SYNAPSE_REST_API);
+            }
             if (serviceName != null && synCfg.getProxyService(serviceName) != null) {
```
isudana: I think we can arrange these checks in a way that we don't have to check whether service is a proxy or an api two times. For an example we can nest synCfg.getProxyService(serviceName) with serviceName != null

PR #815, opened 2017-09-08

Nirthika:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
@@ -1,4 +1,6 @@
 /*
+ *  Copyright (c) 2005-2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
  *  Licensed to the Apache Software Foundation (ASF) under one
  *  or more contributor license agreements.  See the NOTICE file
  *  distributed with this work for additional information
@@ -34,6 +36,11 @@ import org.apache.synapse.mediators.elementary.Source;
 import org.apache.synapse.mediators.elementary.Target;
 
 import java.util.Properties;
+
+/**
+ * Factory for {@link EnrichMediator} instances.
+ *
+ */
 
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
     private static final Log logger = LogFactory.getLog(EnrichMediatorFactory.class.getName());
@@ -200,34 +207,45 @@ public class EnrichMediatorFactory extends AbstractMediatorFactory {
         return XML_Q;
     }
 
-    private void validateTypeCombination(OMElement sourceEle, OMElement targetEle) {
+    /**
+     *
+     * @param sourceElement
+     * @param targetElement
+     *
+     */
+
+    private void validateTypeCombination(OMElement sourceElement, OMElement targetElement) {
         int sourceType = -1;
         int targetType = -1;
 
         // source type attribute
-        OMAttribute sourceTypeAttr = sourceEle.getAttribute(ATT_TYPE);
+        OMAttribute sourceTypeAttr = sourceElement.getAttribute(ATT_TYPE);
         if (sourceTypeAttr != null && sourceTypeAttr.getAttributeValue() != null) {
             sourceType = convertTypeToInt(sourceTypeAttr.getAttributeValue());
 
             // Source type is different form the existing (0-custom, 1-envelope, 2-body, 3-property, 4-inline)
             if (sourceType < 0) {
-                logger.info("Un-expected source type");
+                throw new SynapseException("Un-expected source type");
             }
+        } else {
+            throw new SynapseException("Source type attribute can't be null");
         }
 
         // target type attribute
-        OMAttribute targetTypeAttr = targetEle.getAttribute(ATT_TYPE);
+        OMAttribute targetTypeAttr = targetElement.getAttribute(ATT_TYPE);
         if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
             targetType = convertTypeToInt(targetTypeAttr.getAttributeValue());
 
             // check if target type is different form the existing (0-custom, 1-envelope, 2-body, 3-property, 4-inline)
             if (targetType < 0) {
-                logger.info("Un-expected target type");
+                throw new SynapseException("Un-expected target type");
             }
             // check if target type is 4-inline
             if (targetType == 4) {
                 throw new SynapseException("Inline not support for target attribute");
             }
+        } else {
+            throw new SynapseException("Target type attribute can't be null");
         }
         /*
             check the wrong combination such as
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
@@ -21,7 +21,10 @@ package org.apache.synapse.config.xml;
 import org.apache.axiom.om.OMAbstractFactory;
 import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.Mediator;
+import org.apache.synapse.SynapseException;
 import org.jaxen.JaxenException;
 
 import javax.xml.namespace.QName;
@@ -32,8 +35,9 @@ import org.apache.synapse.mediators.elementary.Target;
 
 import java.util.Properties;
 
-
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
+    private static final Log logger = LogFactory.getLog(EnrichMediatorFactory.class.getName());
+
     private static final QName XML_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "enrich");
     private static final QName ATT_PROPERTY = new QName("property");
     private static final QName ATT_XPATH = new QName("xpath");
@@ -53,8 +57,8 @@ public class EnrichMediatorFactory extends AbstractMediatorFactory {
     @Override
     protected Mediator createSpecificMediator(OMElement elem, Properties properties) {
         if (!XML_Q.equals(elem.getQName())) {
-            handleException("Unable to create the enrich mediator. " +
-                    "Unexpected element as the enrich mediator configuration");
+            handleException("Unable to create the enrich mediator. "
+                    + "Unexpected element as the enrich mediator configuration");
         }
 
         EnrichMediator enrich = new EnrichMediator();
@@ -74,6 +78,8 @@ public class EnrichMediatorFactory extends AbstractMediatorFactory {
         Target target = new Target();
         enrich.setTarget(target);
 
+        validateTypeCombination(sourceEle, targetEle);
+
         populateSource(source, sourceEle);
         populateTarget(target, targetEle);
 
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
         return XML_Q;
     }
 
+    private void validateTypeCombination(OMElement sourceEle, OMElement targetEle) {
```
riyafa: Add method javadoc

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -32,8 +35,9 @@
 
 import java.util.Properties;
 
-
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
```
himasha: Add class javadoc

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
...
+        int targetType = -1;
+
+        // source type attribute
+        OMAttribute sourceTypeAttr = sourceEle.getAttribute(ATT_TYPE);
+        if (sourceTypeAttr != null && sourceTypeAttr.getAttributeValue() != null) {
+            sourceType = convertTypeToInt(sourceTypeAttr.getAttributeValue());
+
+            // Source type is different form the existing (0-custom, 1-envelope, 2-body, 3-property, 4-inline)
+            if (sourceType < 0) {
+                logger.info("Un-expected source type");
```
isudana: Shouldn't we throw an exception here rather than silently failing ?

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
...
+        }
+
+        // target type attribute
+        OMAttribute targetTypeAttr = targetEle.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
+            targetType = convertTypeToInt(targetTypeAttr.getAttributeValue());
+
+            // check if target type is different form the existing (0-custom, 1-envelope, 2-body, 3-property, 4-inline)
+            if (targetType < 0) {
+                logger.info("Un-expected target type");
```
isudana: Again, we need to throw an exception.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
         return XML_Q;
     }
 
+    private void validateTypeCombination(OMElement sourceEle, OMElement targetEle) {
+        int sourceType = -1;
+        int targetType = -1;
+
+        // source type attribute
+        OMAttribute sourceTypeAttr = sourceEle.getAttribute(ATT_TYPE);
+        if (sourceTypeAttr != null && sourceTypeAttr.getAttributeValue() != null) {
```
isudana: can sourceTypeAttr being null is a valid use case ? if not we need to throw an error.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
...
+
+            // Source type is different form the existing (0-custom, 1-envelope, 2-body, 3-property, 4-inline)
+            if (sourceType < 0) {
+                logger.info("Un-expected source type");
+            }
+        }
+
+        // target type attribute
+        OMAttribute targetTypeAttr = targetEle.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
```
isudana: can targetTypeAttr being null a valid use case ?, if not we need to throw an exception

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +200,49 @@ public QName getTagQName() {
...
+            if (targetType == 4) {
+                throw new SynapseException("Inline not support for target attribute");
+            }
+        }
+        /*
+            check the wrong combination such as
+            sourceType = 1-envelope and targetType = 0-custom
+            sourceType = 1-envelope and targetType = 2-body
+            sourceType = 2-body and targetType = 2-body
+            sourceType = 0-custom and targetType = 1-envelope
```
isudana: regarding sourceType = custom targetType = envelope and action=child

Can you please check, in above combination adding the soap header block using inline custom soap header block is possible in enrich mediator implementation before this PR ? If yes we have to consider this as a valid case.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -133,20 +138,28 @@ private void populateSource(Source source, OMElement sourceEle) {
...
+            target.setAction("replace");
+        }
+
         if (typeAttr != null && typeAttr.getAttributeValue() != null) {
-            int type = convertTypeToInit(typeAttr.getAttributeValue());
+            int type = convertTypeToInt(typeAttr.getAttributeValue());
             if (type >= 0) {
                 target.setTargetType(type);
+                if (type == 1) {
+                    if (!target.getAction().equals("replace")) {
```
isudana: we may have to change this depending on the result of the test I have mentioned below comment on (sourceType = custom targetType = envelope and action=child)

Nirthika: when we have targetType = envelope, then the source value should be an envelope. Otherwise it should show the following error. "ERROR - EnrichMediator SOAPEnvelope is expected"
If we have targetType = envelope and an source value also envelope, then the output will be an envelope within an envelope. So this is not a valid combination.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: for the above combination 
when we have targetType = envelope, then the source value should be an envelope. Otherwise it should show the following error. "ERROR - EnrichMediator SOAPEnvelope is expected"
If we have targetType = envelope and an source value also envelope, then the output will be an envelope within an envelope. So this is not a valid combination.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -119,8 +132,7 @@ private void populateSource(Source source, OMElement sourceEle) {
 
             if (inlineElem != null) {
                 source.setInlineOMNode(inlineElem);
-            } else if (sourceEle.getText() != null
-                    && (!sourceEle.getText().equals(""))) {
+            } else if (sourceEle.getText() != null && (!sourceEle.getText().equals(""))) {
```
milindaperera: we don't need two comparisons if we do like this:
if (!("").equals(sourceEle.getText()))

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
isudana: ok, thanks.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
ramith: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-

- This also has similar effect ( but may be we don't use StringUtils in this code base)

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +207,60 @@ public QName getTagQName() {
...
     }
 
+    /**
+     *
+     * @param sourceElement
+     * @param targetElement
+     *
+     */
+
+    private void validateTypeCombination(OMElement sourceElement, OMElement targetElement) {
```
ramith: why not have java docs to explain what this does ? ( why it does it in that way etc)

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: `(!("").equals(sourceEle.getText()))` this only check whether sourceEle.getText() is empty or not, but `sourceEle.getText() != null` this will check whether sourceEle.getText() is null or not, so we need two comparisons to check both.

milindaperera: @Nirthika  yes you are correct. thanks for the explanation.
StringUtils option is also can use here **only if we already using** commons-lang within EI. otherwise IMHO what you have done here is correct

Nirthika: @ramith yes we have **commons-lang** dependency but we don't use **StringUtils** in this code base. We can use `!StringUtils.isBlank(sourceEle.getText())` instead of 
`sourceEle.getText() != null && (!sourceEle.getText().equals(""))`

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -32,8 +38,13 @@
 
 import java.util.Properties;
 
+/**
+ * Factory for {@link EnrichMediator} instances.
+ */
```
milindaperera: Let's remove empty new line after all Javadoc comments

vinok88: Those are not introduced by this commit.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: done

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Done

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Done

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Done

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Done

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Done

PR #821, opened 2017-09-21
Introduce changes to preserve tasks when clustered

MaryamZi:
With Quartz's clustering features being introduced in ntask (wso2/carbon-commons/pull/279), we need to ensure that when a node leaves, the relevant tasks remain in the shared database, for other nodes to pick them up (fail-over). With the current architecture, in a graceful shutdown, with the undeployment of artifacts the relevant tasks are deleted, resulting in a deletion from the database.

This PR introduces changes to enable preservation of the relevant tasks, if undeployment is being called during a graceful shutdown.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
@@ -60,6 +60,7 @@ import org.apache.synapse.message.processor.impl.AbstractMessageProcessor;
 import org.apache.synapse.message.store.MessageStore;
 import org.apache.synapse.registry.Registry;
 import org.apache.synapse.rest.API;
+import org.apache.synapse.startup.quartz.StartUpController;
 import org.apache.synapse.task.TaskManager;
 import org.apache.synapse.util.xpath.ext.SynapseXpathFunctionContextProvider;
 import org.apache.synapse.util.xpath.ext.SynapseXpathVariableResolver;
@@ -1403,7 +1404,14 @@ public class SynapseConfiguration implements ManagedLifecycle, SynapseArtifact {
         destroy(false);
     }
 
-    public synchronized void destroy(boolean preserverState) {
+    /**
+     * Method to destroy the stateful managed parts of the configuration preserving state as required and specifying
+     * if destroy is being called in a shut down.
+     *
+     * @param preserveState whether state should be preserved
+     * @param isShuttingDown whether destroy is being called in a shut down
+     */
+    public synchronized void destroy(boolean preserveState, boolean isShuttingDown) {
 
         if (log.isDebugEnabled()) {
             log.debug("Destroying the Synapse Configuration");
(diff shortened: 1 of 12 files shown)


modules/tasks/src/main/java/org/apache/synapse/task/TaskManager.java
```diff
@@ -16,6 +16,15 @@
 
     public boolean delete(String name);
 
+	/**
+	 * Method to delete the task with the specified name, specifying whether or not deleting is being done with the
+	 * server shutting down.
```
Manuri: Add newline

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -293,14 +296,28 @@ public boolean stop() {
 	}
 
 	@Override
+	public boolean stop() {
+    	return stop(false);
```
Manuri: Fix indentation

modules/tasks/src/main/java/org/apache/synapse/task/TaskManager.java
```diff
@@ -117,4 +126,10 @@
      */
     boolean isTaskExist(final String taskName);
 
+	/**
+	 * Method to retieve whether the TaskManager is managing tasks in a cluster.
+	 *
```
himasha: Remove additional line

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
```diff
@@ -1473,14 +1478,18 @@ public synchronized void destroy(boolean preserverState) {
...
-            mp.destroy();
+            ((AbstractMessageProcessor) mp).destroy(preserverState, isShuttingDown);
         }
 
         for (API api : apiTable.values()) {
             api.destroy();
         }
     }
 
+    public synchronized void destroy(boolean preserverState) {
```
sdkottegoda: typo preserver? Here the input parameter is preserveState, but at line 1452 you pass the value of isShuttingDown. There's also another method introduced with both parameters preserveState and isShuttingDown which means that the 2 values could be different. This seems confusing to me. WDYT?

modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -243,8 +243,7 @@ public void setParameters(Map<String, Object> parameters) {
 	}
 
 
-    @Override
-    public boolean stop() {
+    public boolean stop(boolean isShuttingDown) {
```
abeykoon: To me it looks like we need to have two methods
ScheduledMessageProcessor.stop()
ScheduledMessageProcessor.pause()

Then we can do what is necessary without passing isShuttingDown all the way down the methods? Of cause, if this parameter is useful for another, then it is fine. :)

MaryamZi: In general the isShuttingDown parameter was introduced to prevent deleting the Quartz task in a destroy if destroy is called in a graceful shut down, to allow another node to pick it up and execute it. This is required with the new implementation which uses Quartz's clustering features, which is based on sharing a database which holds the task data.

For ScheduledMessageProcessor specifically, the isShuttingDown parameter is used to NOT pause the task, if destroy is called in a graceful shut down, to allow another node to continue execution. If the task is paused in a shut down, another node will not execute it with the new implementation, since the task would be marked as paused in the shared database.

VijithaEkanayake:
Let's Reopen this PR, Once we have finalized the way to address existing limitations.

PR #823, opened 2017-09-22
add jaccoco coverage for unit test

nuwanw:
Once the modules is built with test, the coverage stat can be find in target/coverage-reports directory in each module


modules/coverage-report/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ~ Licensed to the Apache Software Foundation (ASF) under one ~ or more
```
milindaperera: wrong license header

nuwanw: Same license in all the pom file in synapse

milindaperera: I thought since we adding new module we have to use wso2 license header. If having the same License header in pom files is the requirement, then, this looks good. 👍

PR #832, opened 2017-09-30
Fixing APIMANAGER-5916

sajhak:
This PR contains the fix for java.lang.ClassCastException issue, which occurs in a PUT request with an empty body ('{}') and application/octet-stream content type


sajhak:
Need to add commit description. Will close this PR and send a new PR

PR #835, opened 2017-10-01
[ESBJAVA-5058]Add initAxis2ClientOptions,axis2xml and clientRepositor…

jananithangavel:
…y to Call mediator in blocking mode

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/CallMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/CallMediatorFactory.java
@@ -27,6 +27,7 @@ import org.apache.synapse.endpoints.Endpoint;
 import org.apache.synapse.mediators.builtin.CallMediator;
 
 import javax.xml.namespace.QName;
+import java.io.File;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
```diff
@@ -134,6 +140,10 @@ private boolean handleBlockingCall(MessageContext synInCtx) {
         }
 
         MessageContext resultMsgCtx = null;
+        //If initClientOption is False then set it with blockingMsgsender
+        if (!initClientOptions) {
+            blockingMsgSender.setInitClientOptions(false);
```
a5anka: Can we write this as `blockingMsgSender.setInitClientOptions(initClientOptions)`?

modules/core/src/main/java/org/apache/synapse/config/xml/CallMediatorFactory.java
```diff
@@ -76,6 +77,9 @@
     private static final QName CALL_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "call");
     private static final QName ENDPOINT_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "endpoint");
     private static final QName BLOCKING_Q = new QName("blocking");
+    private static final QName ATT_INIT_AXI2_CLIENT_OPTIONS = new QName("initAxis2ClientOptions");
```
a5anka: The letter `S` is missing in the word `AXI2`. Shall we fix?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
jananithangavel: Done

modules/core/src/main/java/org/apache/synapse/config/xml/CallMediatorFactory.java
jananithangavel: Fixed

PR #759, opened 2017-06-28
Fix ESBJAVA-4902 Parsing Json in response

riyafa:

PR #839, opened 2017-10-03
Improve PR #759 based on comments

riyafa:


modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
```diff
@@ -43,35 +46,60 @@
     private static final String failTestVal5 = "-abc";
     private static final String failTestVal6 = "\"Hello world!";
 
+    /**
+     * Tests JSON value enclosed in double quotes
```
sdkottegoda: Doc comments need to end with a period. Shall we correct in all places?

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
```diff
@@ -43,35 +46,60 @@
...
 
+    /**
+     * Tests JSON value number
+     */
     public void testCase6() {
         runTest(numberVal2);
     }
 
+    /**
+     * Tests JSON value number
```
sdkottegoda: What's the difference between this and the method above? Shall we have the doc, in a way that the test is clear?

modules/commons/src/main/java/org/apache/synapse/commons/json/JsonUtil.java
```diff
@@ -547,33 +547,39 @@ public static OMElement getNewJsonPayload(MessageContext messageContext, InputSt
...
                     isValue = false;
-                } else if (c == '"' || c == 't' || c == 'n' || c == 'f' || c == '-' || c == '0' || c == '1' ||
-                        c == '2' || c == '3' || c == '4' || c == '5' || c == '6' || c == '7' ||
-                        c == '8' || c == '9') {
+                } else if (character == '"' || character == 't' || character == 'n' || character == 'f' ||
+                        character == '-' || character == '0' || character == '1' ||
+                        character == '2' || character == '3' || character == '4' || character == '5' ||
+                        character == '6' || character == '7' ||
+                        character == '8' || character ==
+                        '9') { //a value can be a quoted string (") true (t), false (f), null (n) or a number
```
a5anka: breaking an equal condition to two lines is not very readable. Shall we fix it?

PR #858, opened 2017-10-18
Add synapse-commons-json test cases

manoramahp:

### DIFF
--- a/modules/commons/src/test/java/org/apache/synapse/commons/json/JsonDataSourceImplTest.java
+++ b/modules/commons/src/test/java/org/apache/synapse/commons/json/JsonDataSourceImplTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.synapse.commons.json;
+
+import junit.framework.TestCase;
+import org.apache.axiom.om.OMDataSource;
+
+import javax.xml.stream.XMLOutputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamWriter;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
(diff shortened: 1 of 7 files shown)


modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
```diff
@@ -119,4 +120,13 @@ public void runFailTest(String value) {
             assert true;
         }
     }
+
+    public void testWriteAsJsonNullElement() {
+        try {
+            OutputStream out = Util.newOutputStream();
```
Manuri: Should we not close the stream at the end?

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonDataSourceImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+import javax.xml.stream.XMLStreamWriter;
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.StringWriter;
+import java.io.Writer;
+
+public class JsonDataSourceImplTest extends TestCase {
+    public static final String expectedJSON = "{\n" +
+            "    \"id\":\"0001\",\n" +
```
sdkottegoda: Line wrapping should happen before an operator

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonStreamingBuilderTest.java
```diff
@@ -50,9 +66,33 @@ public void testInvalidJson() {
...
         }
     }
+
+    public void testProcessDocumentInputStreamNull() {
+
+        try {
+            MessageContext messageContext = Util.newMessageContext();
+            Builder jsonBuilder = Util.newJsonStreamBuilder();
+            jsonBuilder.processDocument(null, "application/json", messageContext);
+            assertFalse(true);
```
nuwanw: use Assert.fail instead of assertFalse(true);

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
```diff
@@ -119,4 +120,13 @@ public void runFailTest(String value) {
...
+            JsonUtil.writeAsJson((OMElement) null, out);
```
nuwanw: there should be a Assert.fail() statement to identify if the exception not thrown

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonDataSourceImplTest.java
manoramahp: Done

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
manoramahp: Done

modules/commons/src/test/java/org/apache/synapse/commons/json/JsonValueTest.java
manoramahp: Done

PR #863, opened 2017-10-20
Unit tests for synapse TaskScheduler

hastef88:

### DIFF
--- a/modules/tasks/pom.xml
+++ b/modules/tasks/pom.xml
@@ -137,11 +137,6 @@
             <artifactId>powermock-module-junit4</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.synapse</groupId>
-            <artifactId>synapse-core</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <properties />
--- a/modules/tasks/pom.xml
+++ b/modules/tasks/pom.xml
@@ -137,6 +137,11 @@
             <artifactId>powermock-module-junit4</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.synapse</groupId>
+            <artifactId>synapse-core</artifactId>
+            <scope>test</scope>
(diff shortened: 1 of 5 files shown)


modules/tasks/src/test/java/org/apache/synapse/task/service/TestTaskScheduler.java
```diff
@@ -0,0 +1,89 @@
...
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+
+import javax.xml.stream.XMLStreamException;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.util.Properties;
+
+@PrepareForTest(QuartzTaskManager.class)
+public class TestTaskScheduler {
```
nuwanw: rename the class the TaskScheduler to keep the consistence as other test classes

modules/tasks/pom.xml
```diff
@@ -137,6 +137,11 @@
             <artifactId>powermock-module-junit4</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
nuwanw: can not get synapse core as a dependency, task is building before the core

modules/tasks/src/test/java/org/apache/synapse/task/service/TestTaskScheduler.java
hastef88: Corrected.

modules/tasks/pom.xml
hastef88: (+1) Mocked the TaskManager interface instead of the quartz implementation.

modules/tasks/src/test/resources/task/task_TestScheduler.xml
```diff
@@ -0,0 +1,14 @@
+<task class="org.apache.synapse.task.service.sample.CustomTask"
```
srsharon: license on this file is missing :)

PR #870, opened 2017-10-20
Add unit test for TaskDescriptionFactory

nuwanw:

### DIFF
--- a/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionTest.java
+++ b/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionTest.java
@@ -1,0 +1,65 @@
+/*
+*  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
+*  WSO2 Inc. licenses this file to you under the Apache License,
+*  Version 2.0 (the "License"); you may not use this file except
+*  in compliance with the License.
+*  You may obtain a copy of the License at
+*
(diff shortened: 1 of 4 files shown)


modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionTest.java
```diff
@@ -0,0 +1,65 @@
...
+        properties.put("format", "soap11");
+        taskDescription.setProperties(properties);
+        Assert.assertEquals("Property mismatched", properties, taskDescription.getProperties());
+        Assert.assertEquals("soap11", taskDescription.getProperty("format"));
+        taskDescription.setProperties(null);
+    }
+
+    @Test()
+    public void getTaskImplClassNameTest() {
+        final String classNameBySetter = "org.wso2.task.MyTaskImpl1";
```
chanikag: The constant (variable) name shouldn't be in block capitals?

nuwanw: since it is inside method, the name is same as other variables

modules/tasks/src/test/resources/task/task.xml
```diff
@@ -0,0 +1,14 @@
+<task class="org.apache.synapse.task.impl.CustomTaskTestImpl"
```
srsharon: Shall we add the license here :)

srsharon: License needs to be added on this file

PR #872, opened 2017-10-21
Tests

riyafa:


a5anka:
Can you put only your commit in the PR? It seems some other commits from the master branch is also included.

PR #873, opened 2017-10-22
Modify property mediator to remove properties from axis2-client scope

ThishaniLucas:
The properties that are set in the axis2-client scope are not getting removed. This is the fix for that issue.
 https://github.com/wso2/product-ei/issues/1193

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
@@ -214,8 +214,7 @@ public class PropertyMediator extends AbstractMediator {
                     pros.remove(name);
                 }
 
-            } else if ((XMLConfigConstants.SCOPE_AXIS2.equals(scope) ||
-                XMLConfigConstants.SCOPE_CLIENT.equals(scope))
+            } else if (XMLConfigConstants.SCOPE_AXIS2.equals(scope)
                 && synCtx instanceof Axis2MessageContext) {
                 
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
```diff
@@ -224,6 +223,15 @@ public boolean mediate(MessageContext synCtx) {
...
                 axis2MessageCtx.removeProperty(name);
 
+            } else if (XMLConfigConstants.SCOPE_CLIENT.equals(scope)
+                    && synCtx instanceof Axis2MessageContext) {
+
+                //Removing property from the Axis2-client Message Context
+                Axis2MessageContext axis2smc = (Axis2MessageContext) synCtx;
+                org.apache.axis2.context.MessageContext axis2MessageCtx =
+                        axis2smc.getAxis2MessageContext();
+                axis2MessageCtx.getOptions().setProperty(name, null);
```
nuwanw: why the value is null

ThishaniLucas: We don't have a remove-property option at axis2MessageCtx.getOptions(). Therefore, in order to remove the property, I'm setting its value to null.

nuwanw: can you add it as a code comment

ThishaniLucas: done

PR #877, opened 2017-10-23
Add passthru-http.properties file

riyafa:


repository/conf/passthru-http.properties
```diff
@@ -0,0 +1,47 @@
+#
+# Copyright (c) 2016, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
srsharon: the year of the license needs to be changed

PR #880, opened 2017-10-23
add test class for TaskDescriptionSerializer

nuwanw:

### DIFF
--- a/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
+++ b/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
@@ -30,8 +30,8 @@ import java.io.FileNotFoundException;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
 
-/*
-Test class for TaskDescriptionSerializer.
+/**
+ * Test class for TaskDescriptionSerializer.
  */
 public class TaskDescriptionSerializerTest {
     private final String SYNAPSE_NAMESPACE = TaskTestConstant.SYNAPSE_NAMESPACE;
--- a/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
+++ b/modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
@@ -1,0 +1,58 @@
+/*
+*  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
+*  WSO2 Inc. licenses this file to you under the Apache License,
+*  Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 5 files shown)


modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.xml.stream.XMLStreamException;
+
+/*
+Test class for TaskDescriptionSerializer.
+ */
+public class TaskDescriptionSerializerTest {
+    private final String SYNAPSE_NAMESPACE = TaskTestConstant.SYNAPSE_NAMESPACE;
+    private final OMNamespace SYNAPSE_OMNAMESPACE = TaskTestConstant.SYNAPSE_OMNAMESPACE;
+
+    @Test()
```
Asitha: Shall we have a doc comment for the test? Maybe explaining the test steps?

modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Test;
+
+import java.io.FileNotFoundException;
+import javax.xml.namespace.QName;
+import javax.xml.stream.XMLStreamException;
+
+/*
+Test class for TaskDescriptionSerializer.
+ */
+public class TaskDescriptionSerializerTest {
```
Asitha: Shall we update the doc comment  as follows?
```
/*
 * Test class for TaskDescriptionSerializer.
 */
```

a5anka: AFAIK Javadoc comment should start with `/**`. If we use `/*`, it will be treated as a multiline comment.

modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
nuwanw: for unit test method, we do not need to have the java doc comments.

modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
nuwanw: Done

modules/tasks/src/test/java/org/apache/synapse/task/TaskDescriptionSerializerTest.java
nuwanw: We can add the brief summery of the test

Asitha: +1

PR #882, opened 2017-10-25
add a test class for PassThroughHttpListenerTest

nuwanw:

### DIFF
--- a/modules/transports/core/nhttp/pom.xml
+++ b/modules/transports/core/nhttp/pom.xml
@@ -207,6 +207,16 @@
             <classifier>runtime</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <properties />
--- a/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
+++ b/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
@@ -1,0 +1,50 @@
+/*
+*  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
+*  WSO2 Inc. licenses this file to you under the Apache License,
+*  Version 2.0 (the "License"); you may not use this file except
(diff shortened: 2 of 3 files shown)


modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
```diff
@@ -0,0 +1,50 @@
...
+            socket = new Socket(host, port);
+            isPortOpen = socket.isConnected();
+        } catch (IOException e) {
+            isPortOpen = false;
+        } finally {
+            try {
+                if ((socket != null) && (socket.isConnected())) {
+                    socket.close();
+                }
+            } catch (IOException e) {
```
a5anka: Should we catch this exception since this is a test util method?

modules/transports/core/nhttp/pom.xml
```diff
@@ -207,6 +207,16 @@
             <classifier>runtime</classifier>
```
srsharon: Can you update the license type.

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
nuwanw: Yes. The method return a boolean value for assertions.

a5anka: :+1:

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
```diff
@@ -0,0 +1,50 @@
...
+                //                log.error("Can not close the socket with is used to check the server status ", e);
```
a5anka: Shall we remove the commented out code?

modules/transports/core/nhttp/pom.xml
```diff
@@ -207,6 +207,16 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
```
a5anka: Have we defined the dependency versions in parent? Just checking since it not done in this PR.

nuwanw: Yes. It is already there

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/TCPUtils.java
nuwanw: Done

PR #885, opened 2017-10-26
Add tests for pass-through http sender

ThishaniLucas:

### DIFF
--- a/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
+++ b/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
@@ -1,0 +1,133 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.synapse.transport.passthru;
+
+import junit.framework.Assert;
+import junit.framework.TestCase;
+import org.apache.axis2.Constants;
+import org.apache.axis2.context.ConfigurationContext;
+import org.apache.axis2.context.MessageContext;
+import org.apache.axis2.context.OperationContext;
+import org.apache.axis2.description.TransportOutDescription;
+import org.apache.axis2.engine.AxisConfiguration;
+import org.apache.axis2.engine.Handler;
+import org.apache.axis2.transport.base.threads.NativeWorkerPool;
+import org.apache.axis2.transport.base.threads.WorkerPool;
+import org.apache.commons.logging.Log;
+import org.apache.http.impl.nio.DefaultNHttpServerConnection;
+import org.apache.http.nio.NHttpServerConnection;
+import org.apache.http.nio.reactor.IOSession;
+import org.apache.synapse.transport.nhttp.NhttpConstants;
+import org.apache.synapse.transport.passthru.config.TargetConfiguration;
+import org.apache.synapse.transport.passthru.util.BufferFactory;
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
```diff
@@ -0,0 +1,133 @@
...
+        messageContext.setOperationContext(new OperationContext());
+
+        sender = PowerMockito.spy(sender);
+
+        PowerMockito.when(targetConfiguration.getBufferFactory()).thenReturn(factory);
+        PowerMockito.doNothing().when(sender, "sendRequestContent", any(MessageContext.class));
+
+        Handler.InvocationResponse response = sender.invoke(messageContext);
+
+        Assert.assertNotNull(response);
```
manoramahp: Shall we add an assert message here?

ThishaniLucas: done

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.powermock.modules.junit4.PowerMockRunner;
+import org.wso2.caching.digest.DigestGenerator;
+import java.util.HashMap;
+import java.util.Map;
+import static org.mockito.ArgumentMatchers.any;
+
+@RunWith(PowerMockRunner.class)
+@PowerMockIgnore("javax.management.*")
+@PrepareForTest({ PassThroughHttpSender.class, SourceContext.class, SourceResponseFactory.class})
+public class PassThroughHttpSenderTest extends TestCase {
```
ramith: class comments?

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
```diff
@@ -0,0 +1,133 @@
...
+        configurationContext.setProperty(PassThroughConstants.PASS_THROUGH_TRANSPORT_WORKER_POOL, workerPool);
+        TransportOutDescription transportOutDescription = new TransportOutDescription("passthru");
+
+        PassThroughHttpSender passThroughHttpSender = new PassThroughHttpSender();
+
+        passThroughHttpSender.init(configurationContext, transportOutDescription);
+    }
+
+    @Test
+    public void testInvoke() throws Exception {
```
ramith: Documentation on testcase?

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        Map headers = new HashMap();
+        headers.put("randomKey", "randomValue");
+
+        messageContext.setProperty(MessageContext.TRANSPORT_HEADERS, headers);
+
+        sender = PowerMockito.spy(sender);
+
+        PowerMockito.when(digestGenerator.getDigest(any(MessageContext.class))).thenReturn("testString");
+        sender.submitResponse(messageContext);
```
ramith: any reason why we don't have a assert in this test? (if its a testcase what are we testing)

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
ThishaniLucas: fixed.
https://github.com/wso2/wso2-synapse/pull/893/files

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
ThishaniLucas: fixed
https://github.com/wso2/wso2-synapse/pull/893/files

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/PassThroughHttpSenderTest.java
ThishaniLucas: The submit-response method does not return anything but print some logs. The method completes only if there are no errors. Therefore, it's assumed that if the method completes, the test case passes.  Otherwise we'll have to have a log-viewer client as in the integration tests and check for logs.

PR #887, opened 2017-10-26
add test class for passthru SourceHandler

nuwanw:

### DIFF
--- a/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/SourceHandlerTest.java
+++ b/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/SourceHandlerTest.java
@@ -1,0 +1,95 @@
+/*
+*  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/SourceHandlerTest.java
```diff
@@ -0,0 +1,95 @@
...
+        ConfigurationContext cfgCtx = new ConfigurationContext(new AxisConfiguration());
+        cfgCtx.setServicePath("services");
+        cfgCtx.setContextRoot("/");
+        passThroughHttpListener.init(cfgCtx, transportInDescription);
+        passThroughHttpListener.start();
+        Assert.assertTrue("Listener port not open", TCPUtils.isPortOpen(PORT, HOST));
+    }
+
+    @Test
+    public void testRequestAck() throws Exception {
```
milindaperera: shall we add description what is tested in this test case

nuwanw: will add later

PR #899, opened 2017-10-30
Move private methods to utility class

nuwanw:

### DIFF
--- a/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/ServiceUtils.java
+++ b/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/ServiceUtils.java
@@ -29,6 +29,9 @@ import org.apache.synapse.transport.passthru.PassThroughHttpSender;
 
 import java.io.IOException;
 
+/**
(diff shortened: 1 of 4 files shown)


modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/utils/ServiceUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.axis2.context.ServiceContext;
+import org.apache.axis2.description.InOutAxisOperation;
+import org.apache.axis2.description.TransportOutDescription;
+import org.apache.axis2.engine.AxisConfiguration;
+import org.apache.http.HttpException;
+import org.apache.synapse.transport.passthru.PassThroughHttpSender;
+
+import java.io.IOException;
+
+public class ServiceUtils {
```
milindaperera: Shall we add class Javadoc

nuwanw: Done

PR #902, opened 2017-10-31
Add unit tests for RabbitMQStore class

GDLMadushanka:
## Automation tests
 - Unit tests 
   > 24%

### DIFF
--- a/modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
+++ b/modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
@@ -1,0 +1,156 @@
+/**
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * <p>
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
```diff
@@ -0,0 +1,154 @@
...
+        Assert.assertEquals("Comparing MessageConsumer id", "[null-C-1]", messageConsumer.getId());
+    }
+
+    /**
+     * Create axis2MessageContext
+     */
+    @Test
+    public void testNewAxis2Mc() {
+        MessageContext messageContext = rabbitMQStore.newAxis2Mc();
+        Assert.assertEquals(messageContext.getConfigurationContext(), configurationContext);
```
chanikag: Add an Assert message here?

modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Test
+    public void testRemove() {
+        Assert.assertNull(rabbitMQStore.remove());
```
chanikag: Add an Assert message here?

modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    @Test
+    public void testNewAxis2Mc() {
+        MessageContext messageContext = rabbitMQStore.newAxis2Mc();
+        Assert.assertEquals("MessageContext should be initiated with configurationContext",messageContext
+                .getConfigurationContext(), configurationContext);
+    }
+
+    //Following test cases assert currently not implemented methods which are just returning null
+    @Test
```
pamod: Describe the following test cases

GDLMadushanka: Methods tested in these tests are not implemented yet. I add these test case for coverage purposes and for the ease of changing them when methods are actually implemented.

PR #903, opened 2017-10-31
Add tests for pass-through transport response

ThishaniLucas:

### DIFF
--- a/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
+++ b/modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
@@ -1,0 +1,148 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.synapse.transport.passthru;
+
+import junit.framework.Assert;
+import junit.framework.TestCase;
+import org.apache.axis2.context.ConfigurationContext;
+import org.apache.axis2.engine.AxisConfiguration;
+import org.apache.axis2.transport.base.threads.NativeWorkerPool;
+import org.apache.axis2.transport.base.threads.WorkerPool;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpResponse;
+import org.apache.http.nio.ContentDecoder;
+import org.apache.http.nio.NHttpClientConnection;
+import org.apache.synapse.transport.passthru.config.TargetConfiguration;
+import org.apache.synapse.transport.passthru.connections.TargetConnections;
+import org.apache.synapse.transport.passthru.jmx.PassThroughTransportMetricsCollector;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import java.nio.ByteBuffer;
+
+import static org.mockito.ArgumentMatchers.any;
+
+/**
+ * Test class for TargetResponse
+ */
+@RunWith(PowerMockRunner.class)
+@PowerMockIgnore("javax.management.*")
+@PrepareForTest({ TargetContext.class })
+public class TargetResponseTest extends TestCase {
+    private static Log logger = LogFactory.getLog(TargetResponseTest.class.getName());
+
+    /**
+     * Testing the starting of target response when response body is not expected
+     *
+     * @throws Exception
+     */
+    @Test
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+import static org.mockito.ArgumentMatchers.any;
+
+/**
+ * Test class for TargetResponse
+ */
+@RunWith(PowerMockRunner.class)
+@PowerMockIgnore("javax.management.*")
+@PrepareForTest({ TargetContext.class})
+public class TargetResponseTest extends TestCase{
```
nuwanw: Code need to be refactored. space required before the {

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+        TargetResponse targetResponse = new TargetResponse(targetConfiguration, response, conn, false, false);
+
+        try{
+            targetResponse.start(conn);
+        } catch (Exception e){
+            logger.error(e);
+            Assert.fail("Unable to start the target response!");
+        }
+
```
nuwanw: no need of new lines

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,142 @@
...
+        PowerMockito.when(TargetContext.get(any(NHttpClientConnection.class))).thenReturn(cntxt);
+        PowerMockito.when(decoder.read(any(ByteBuffer.class))).thenReturn(12);
+        PowerMockito.when(decoder.isCompleted()).thenReturn(true);
+
+        TargetResponse targetResponse = new TargetResponse(targetConfiguration, response, conn, true, false);
+        targetResponse.start(conn);
+        int result = targetResponse.read(conn, decoder);
+
+        Assert.assertTrue("Couldn't read from the pipe!", result==12);
+
```
manoramahp: We can use assertequals() method here

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
ThishaniLucas: done

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
ThishaniLucas: fixed

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
ThishaniLucas: done

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,148 @@
...
+        NHttpClientConnection conn = PowerMockito.mock(NHttpClientConnection.class, Mockito.RETURNS_DEEP_STUBS);
+        ContentDecoder decoder = PowerMockito.mock(ContentDecoder.class);
+        TargetConnections connections = PowerMockito.mock(TargetConnections.class);
+        targetConfiguration.setConnections(connections);
+
+        PowerMockito.mockStatic(TargetContext.class);
+
+        TargetContext cntxt = new TargetContext(targetConfiguration);
+        PowerMockito.when(TargetContext.get(any(NHttpClientConnection.class))).thenReturn(cntxt);
+        PowerMockito.when(decoder.read(any(ByteBuffer.class))).thenReturn(12);
```
indikasampath2000: What is the magic number 12 here? Can you create a local variable with the meaningful name and use it in the logic?

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        PowerMockito.mockStatic(TargetContext.class);
+
+        TargetResponse targetResponse = new TargetResponse(targetConfiguration, response, conn, false, false);
+
+        try {
+            targetResponse.start(conn);
+        } catch (Exception e) {
+            logger.error(e);
+            Assert.fail("Unable to start the target response!");
```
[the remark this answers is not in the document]
ThishaniLucas: Here we're not actually testing the failure case. The purpose of this fail is to fail the test case in case the start was not successful.

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
```diff
@@ -0,0 +1,148 @@
...
+        TargetContext cntxt = new TargetContext(targetConfiguration);
+        PowerMockito.when(TargetContext.get(any(NHttpClientConnection.class))).thenReturn(cntxt);
+
+        TargetResponse targetResponse = new TargetResponse(targetConfiguration, response, conn, true, false);
+
+        try {
+            targetResponse.start(conn);
+        } catch (Exception e) {
+            logger.error(e);
+            Assert.fail("Unable to start the target response!");
```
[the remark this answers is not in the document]
ThishaniLucas: This test checks whether a connection with hard-coded params can be established to start reading from the target. Since there's no real target, we don't actually get any response. Therefore, we cannot assert based on getStatus.

modules/transports/core/nhttp/src/test/java/org/apache/synapse/transport/passthru/TargetResponseTest.java
ThishaniLucas: It's actually a random value. It's the value that is supposed to be returned by power mockito when a byte buffer is read.

PR #906, opened 2017-11-01
Add unit tests for BeanMediator 

GDLMadushanka:
## Automation tests
 - Unit tests 
   > 42 %

### DIFF
--- a/modules/core/src/test/java/org/apache/synapse/mediators/bean/BeanMediatorTest.java
+++ b/modules/core/src/test/java/org/apache/synapse/mediators/bean/BeanMediatorTest.java
@@ -1,0 +1,140 @@
+/**
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * <p>
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/synapse/mediators/bean/BeanMediatorTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.synapse.mediators.Value;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+/**
+ * Unit tests for MeanMediator class
+ */
```
manoramahp: Typo; The class name in the comment should be BeanMediator

modules/core/src/test/java/org/apache/synapse/mediators/bean/BeanMediatorTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     * Asserting exception for creating an erroneous Target
+     */
+    @Test
+    public void testErroneousValue() {
+        try {
+            Target target = new Target(ERROR_ATTRIBUTE, TestUtils.createOMElement(XML));
+        } catch (Exception ex) {
```
manoramahp: Add Assert.fail() here

PR #907, opened 2017-11-01
add test case for passthru request and response

nuwanw:
1) Send a request to server
 2) Server send the request to backend
 3) Server respond to client


milindaperera:
Is it possible to locate SimpleHttpServer.java in a more common place to reused in unit tests in other module?

nuwanw:
There is no common place for synapse unit test and this will be used in transport related test. So lets keep it within the test module

PR #912, opened 2017-11-01

Nirthika:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
@@ -1,4 +1,6 @@
 /*
+ *  Copyright (c) 2005-2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
  *  Licensed to the Apache Software Foundation (ASF) under one
  *  or more contributor license agreements.  See the NOTICE file
  *  distributed with this work for additional information
@@ -21,7 +23,11 @@ package org.apache.synapse.config.xml;
 import org.apache.axiom.om.OMAbstractFactory;
 import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.Mediator;
+import org.apache.synapse.SynapseException;
 import org.jaxen.JaxenException;
 
 import javax.xml.namespace.QName;
@@ -32,8 +38,12 @@ import org.apache.synapse.mediators.elementary.Target;
 
 import java.util.Properties;
 
-
+/**
+ * Factory for {@link EnrichMediator} instances.
+ */
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
+    private static final Log logger = LogFactory.getLog(EnrichMediatorFactory.class.getName());
+
     private static final QName XML_Q = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "enrich");
     private static final QName ATT_PROPERTY = new QName("property");
     private static final QName ATT_XPATH = new QName("xpath");
@@ -53,8 +63,8 @@ public class EnrichMediatorFactory extends AbstractMediatorFactory {
     @Override
     protected Mediator createSpecificMediator(OMElement elem, Properties properties) {
         if (!XML_Q.equals(elem.getQName())) {
-            handleException("Unable to create the enrich mediator. " +
-                    "Unexpected element as the enrich mediator configuration");
+            handleException("Unable to create the enrich mediator. "
+                    + "Unexpected element as the enrich mediator configuration");
         }
 
         EnrichMediator enrich = new EnrichMediator();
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -32,8 +38,12 @@
 
 import java.util.Properties;
 
-
+/**
+ * Factory for {@link EnrichMediator} instances.
+ */
 public class EnrichMediatorFactory extends AbstractMediatorFactory {
+    private static final Log logger = LogFactory.getLog(EnrichMediatorFactory.class.getName());
```
VijithaEkanayake: What is the usage of logger variable? if there is no usage shall we remove that variable?

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +206,62 @@ public QName getTagQName() {
...
+        int targetType = -1;
+
+        // source type attribute
+        OMAttribute sourceTypeAttr = sourceElement.getAttribute(ATT_TYPE);
+        if (sourceTypeAttr != null && sourceTypeAttr.getAttributeValue() != null) {
+            sourceType = convertTypeToInt(sourceTypeAttr.getAttributeValue());
+
+            // check if source type is different form the existing
+            if (sourceType < 0) {
+                throw new SynapseException("Un-expected source type");
```
VijithaEkanayake: Shouldn't Unexpected be a single word?

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +206,62 @@ public QName getTagQName() {
...
+            }
+        }
+        // target type attribute
+        OMAttribute targetTypeAttr = targetElement.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
+            targetType = convertTypeToInt(targetTypeAttr.getAttributeValue());
+
+            // check if target type is different form the existing
+            if (targetType < 0) {
+                throw new SynapseException("Un-expected target type");
```
VijithaEkanayake: Same as above

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -133,20 +141,28 @@ private void populateSource(Source source, OMElement sourceEle) {
...
+            target.setAction("replace");
+        }
+
         if (typeAttr != null && typeAttr.getAttributeValue() != null) {
-            int type = convertTypeToInit(typeAttr.getAttributeValue());
+            int type = convertTypeToInt(typeAttr.getAttributeValue());
             if (type >= 0) {
                 target.setTargetType(type);
+                if (type == 1) {
+                    if (!target.getAction().equals("replace")) {
```
milindaperera: Let's merge two if conditions to one

And also please use EnrichMediator.BODY etc instead of using integer values for comparison

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +203,62 @@ public QName getTagQName() {
...
+        OMAttribute targetTypeAttr = targetElement.getAttribute(ATT_TYPE);
+        if (targetTypeAttr != null && targetTypeAttr.getAttributeValue() != null) {
+            targetType = convertTypeToInt(targetTypeAttr.getAttributeValue());
+
+            // check if target type is different form the existing
+            if (targetType < 0) {
+                throw new SynapseException("Unexpected target type");
+            }
+            // check if target type is 4-inline
+            if (targetType == 4) {
```
milindaperera: let's use defined constant for this targetType comparison instead of using integer values

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -187,4 +203,62 @@ public QName getTagQName() {
...
+            sourceType = 1-envelope and targetType = 0-custom
+            sourceType = 1-envelope and targetType = 2-body
+            sourceType = 2-body and targetType = 2-body
+            sourceType = 0-custom and targetType = 1-envelope
+            sourceType = 1-envelope and targetType = 1-envelope
+            sourceType = 2-body and targetType = 1-envelope
+
+         */
+        if (sourceType == 1) {
+            if (targetType == 0 || targetType == 1 || targetType == 2) {
```
milindaperera: here too. use defined constants if available, 
define if not available

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: For `if (type >= 0)` these kind of checking I need integers. So as used in old implementation I use integers here.

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Same as above

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
Nirthika: Same as above

PR #914, opened 2017-11-02
Add tests for JMS store

ThishaniLucas:

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -381,6 +381,16 @@
 			<artifactId>powermock-module-junit4</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.activemq</groupId>
+			<artifactId>activemq-broker</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 <properties>
 		<version.orbit.rabbitmq>3.6.6.wso2v1</version.orbit.rabbitmq>
--- a/modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
+++ b/modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
@@ -1,0 +1,94 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.apache.synapse.message.store.impl.jms;
+
+import junit.framework.Assert;
+import org.apache.axiom.om.OMElement;
+import org.apache.axiom.om.util.AXIOMUtil;
+import org.apache.axiom.soap.SOAPEnvelope;
+import org.apache.axiom.soap.impl.llom.soap11.SOAP11Factory;
+import org.apache.axis2.context.ConfigurationContext;
+import org.apache.synapse.MessageContext;
+import org.apache.synapse.config.SynapseConfiguration;
+import org.apache.synapse.core.SynapseEnvironment;
+import org.apache.synapse.core.axis2.Axis2MessageContext;
+import org.apache.synapse.core.axis2.Axis2SynapseEnvironment;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.apache.activemq.broker.BrokerService;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Test class for Jms Store implementation
+ */
+public class JmsMessageStoreTest{
+
+    private static BrokerService broker;
+
+    @BeforeClass
+    public static void startBroker() throws Exception {
+        broker = new BrokerService();
+        broker.setPersistent(false);
+        broker.addConnector("tcp://127.0.0.1:61616");
+        broker.start();
+        if (!broker.isStarted()) {
+            throw new Exception("Couldn't start the broker!");
+        }
+    }
+
+    /**
+     * Testing whether jms producer is storing the message to the store
+     * @throws Exception
+     */
+    @Test
+    public void testProducer() throws Exception{
+
+        JmsStore jmsStore = new JmsStore();
+
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("java.naming.factory.initial", "org.apache.activemq.jndi.ActiveMQInitialContextFactory");
+        parameters.put("java.naming.provider.url", "tcp://127.0.0.1:61616");
+        jmsStore.setParameters(parameters);
+        jmsStore.setName("TestJmsStore");
+        SynapseConfiguration synapseConfiguration = new SynapseConfiguration();
+        synapseConfiguration.addMessageStore("JMSStore", jmsStore);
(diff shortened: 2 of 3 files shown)


modules/core/pom.xml
```diff
@@ -381,6 +381,18 @@
...
+		<dependency>
+			<groupId>org.apache.activemq</groupId>
+			<artifactId>activemq-broker</artifactId>
+			<version>5.9.1</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+			<version>1.5.2</version>
```
manoramahp: Shall we add these dependencies to the parent pom, so that we can define the versions there

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        Assert.assertNotNull("Could not consume message from the store!", responseContext);
+    }
+
+    @AfterClass
+    public static void stopBroker() throws Exception{
+        if(broker.isStarted()){
+            broker.stop();
+        }
+        System.out.println("Broker Stopped!");
```
chanikag: Shall we log instead of System.out.println

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    private static BrokerService broker;
+
+    @BeforeClass
+    public static void startBroker() throws Exception {
+        broker = new BrokerService();
+        broker.setPersistent(false);
+        broker.addConnector("tcp://127.0.0.1:61616");
+        broker.start();
+        System.out.println("Broker Started!");
```
chanikag: Shall we log instead of System.out.println

modules/core/pom.xml
ThishaniLucas: done

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
ThishaniLucas: implemented a different logic

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
ThishaniLucas: implemented a different logic

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Test class for Jms Store implementation
+ */
+public class JmsMessageStoreTest {
+
+    private static BrokerService broker;
+    private static final Log log = LogFactory.getLog(JmsMessageStoreTest.class);
```
chanikag: As log is not used, shall we remove this and the imports

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.Test;
+import org.apache.activemq.broker.BrokerService;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Test class for Jms Store implementation
+ */
+public class JmsMessageStoreTest {
```
ramith: shouldn't we say this activeMQ jms test case or something and add a test case for wso2 mb as well?

ThishaniLucas: In this test case, we're using an embedded activemq broker. The broker is not running as a separate process. But with wso2 mb, we have to start the broker separately. Anyhow we can try that with andes. But this test case tests only the jms store implementation.

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
ThishaniLucas: done

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
```diff
@@ -0,0 +1,135 @@
...
+        MessageContext messageContext = axis2MessageContext;
+        SOAP11Factory factory = new SOAP11Factory();
+        SOAPEnvelope envelope = factory.createSOAPEnvelope();
+        messageContext.setEnvelope(envelope);
+
+        jmsStore.init(synapseEnvironment);
+
+        JmsProducer jmsProducer = (JmsProducer) jmsStore.getProducer();
+        boolean response = jmsProducer.storeMessage(messageContext);
+
```
nuwanw: need an assertion by checking the message in broker as well

ThishaniLucas: done

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
```diff
@@ -126,6 +126,70 @@ public MessageContext receive() {
...
     }
 
+    public MessageContext receive(long milliSeconds) {
+        boolean error;
+        JMSException exception;
+        if (!checkConnection()) {
+            if (!reconnect()) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug(getId()
+                            + " cannot receive message from store. Cannot reconnect.");
```
sdkottegoda: Can we merge this to the previous line?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
```diff
@@ -126,6 +126,70 @@ public MessageContext receive() {
...
+            return null;
+        }
+        try {
+            Message message = consumer.receive(milliSeconds);
+            if (message == null) {
+                return null;
+            }
+            if (!(message instanceof ObjectMessage)) {
+                logger.warn(getId() + ". Did not receive a javax.jms.ObjectMessage");
+                message.acknowledge(); // TODO:
```
sdkottegoda: Why do we have a TODO here?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
```diff
@@ -126,6 +126,70 @@ public MessageContext receive() {
...
+            }
+            return synapseMc;
+        } catch (JMSException e) {
+            error = true;
+            exception = e;
+        }
+        if (error) {
+            if (!isReceiveError) {
+                logger.error(getId() + " cannot receive message from store. Error:"
+                        + exception.getLocalizedMessage());//, exception);
```
sdkottegoda: Shall we remove the comment?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
ThishaniLucas: done

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
ThishaniLucas: fixed

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
ThishaniLucas: done

modules/core/src/test/java/org/apache/synapse/message/store/impl/jms/JmsMessageStoreTest.java
sdkottegoda: @Ramith, by default jms test cases are written to run with activeMQ. We haven't named the test cases as such in any of the existing test cases.
@ThishaniLucas if we are to add a test case for wso2mb for this tests-platform -> tests-wso2mb is the place for that.

PR #920, opened 2017-11-07
Change surefire version and fix unit tests

GDLMadushanka:

### DIFF
--- a/modules/commons/src/test/java/org/apache/synapse/commons/staxon/core/event/SimpleXMLEventFactoryTest.java
+++ b/modules/commons/src/test/java/org/apache/synapse/commons/staxon/core/event/SimpleXMLEventFactoryTest.java
@@ -236,6 +236,5 @@ public class SimpleXMLEventFactoryTest {
     @Test(expected = UnsupportedOperationException.class)
     public void testCreateDTD() {
         factory.createDTD("foo");
-        Assert.fail("executed without expected exception");
     }
 }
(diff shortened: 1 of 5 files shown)


modules/commons/src/test/java/org/apache/synapse/commons/staxon/core/event/SimpleXMLEventFactoryTest.java
```diff
@@ -235,10 +235,7 @@ public void testCreateProcessingInstruction() throws XMLStreamException {
 
     @Test(expected = UnsupportedOperationException.class)
     public void testCreateDTD() {
-        try {
-            factory.createDTD("foo");
-        } catch (UnsupportedOperationException ex) {
-            //empty
-        }
+        factory.createDTD("foo");
+        Assert.fail("executed without expected exception");
```
nuwanw: we do not need a assert.fail since this statement not executed because of factory.createDTD("foo") throws the exception. It it failed to throw exception, junit will fail the test

modules/commons/src/test/java/org/apache/synapse/commons/staxon/core/json/JsonXMLStreamWriterTest.java
```diff
@@ -259,16 +259,13 @@ public void testMixedValueArray() throws Exception {
...
-        } catch (XMLStreamException ex) {
-            //empty
-        }
+    public void testElementMultipleRoots() throws XMLStreamException {
+        XMLStreamWriter writer = new JsonXMLOutputFactory().createXMLStreamWriter(new StringWriter());
+        writer.writeStartDocument();
+        writer.writeStartElement("foo");
+        writer.writeEndElement();
+        writer.writeStartElement("bar");
+        Assert.fail("execution successful without expected exception");
```
nuwanw: remove assert fail since expected Exception is used

PR #923, opened 2017-11-07
Add unit tests for commons.datasource

GDLMadushanka:
## Automation tests
 - Unit tests 
   > 60%


modules/commons/src/test/java/org/apache/synapse/commons/datasource/DataSourceInformationRepositoryTest.java
```diff
@@ -0,0 +1,195 @@
...
+     */
+    @Test
+    public void testSetRepositoryListenerNull() {
+        try {
+            repository.setRepositoryListener(null);
+            Assert.fail("execution successful without expected exception");
+        } catch (Exception ex) {
+            Assert.assertEquals("asserting exception message", "Provided " +
+                    "DataSourceInformationRepositoryListener instance is null", ex.getMessage());
+            Assert.assertEquals("asserting exception class", SynapseCommonsException.class, ex.getClass());
```
ThishaniLucas: Can't we use expected=SynapseCommonsException.class to test thrown exception? Or expected exception rule?

himasha: +1 , better to use expectedException

GDLMadushanka: expected=SynapseCommonsException.class does not give us a way to assert exception messages.
expected exception rule works same as try catch, except in try catch we have flexibility to use various assert methods
So, do I have to make this change?

himasha: Since exception message is also asserted try/catch should be fine. However JUnit also provides  @Rule annotation where you can assert for exception type + message. Since you are checking for 'SynapseCommonsException' mostly, below approach might be cleaner. 
@Rule
    public SynapseCommonsException thrown = SynapseCommonsException.none();
Then you could add in your test-cases like,
thrown.expect();
thrown.expectMessage();

PR #934, opened 2017-11-08
Add tests for artifact deployment store

ThishaniLucas:


modules/core/src/test/java/org/apache/synapse/deployers/SynapseArtifactDeploymentStoreTest.java
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Test class for SynapseArtifactDeploymentStore
+ */
+public class SynapseArtifactDeploymentStoreTest {
+    /**
+     * Test adding of artifact
+     * @throws Exception
+     */
+    @Test
+    public void testAddArtifact() throws Exception{
```
indikasampath2000: Method signature should format correctly. There should be space before `{`. Please check in the entire PR.

ThishaniLucas: Fixed with https://github.com/wso2/wso2-synapse/pull/1100

PR #937, opened 2017-11-09
Add unit test for TaskManagerFactory and MessgaeProcessorSerializer

Nirothipan:
## Automation tests
 - Unit tests 
   > MessageProcessorSerializer - 88%
  > TaskManagerFactory - 69%

### DIFF
--- a/modules/core/src/test/java/org/apache/synapse/config/xml/MessageProcessorSerializerTest.java
+++ b/modules/core/src/test/java/org/apache/synapse/config/xml/MessageProcessorSerializerTest.java
@@ -1,0 +1,73 @@
+/*
+*  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+*
+*  WSO2 Inc. licenses this file to you under the Apache License,
+*  Version 2.0 (the "License"); you may not use this file except
+*  in compliance with the License.
+*  You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*/
+package org.apache.synapse.config.xml;
+
+import org.apache.axiom.om.OMElement;
+import org.apache.synapse.SynapseException;
+import org.apache.synapse.message.processor.MessageProcessor;
+import org.apache.synapse.message.processor.impl.forwarder.ScheduledMessageForwardingProcessor;
+import org.apache.synapse.message.processor.impl.sampler.SamplingProcessor;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/synapse/config/xml/TaskManagerFactoryTest.java
```diff
@@ -0,0 +1,90 @@
...
+        OMElement element = AXIOMUtil.stringToOM(inputXML);
+        TaskManagerFactory.createTaskManager(element, null);
+    }
+
+    /**
+     * test createTaskManager with provider not specified.
+     *
+     * @throws XMLStreamException - XMLStreamException.
+     */
+    @Test(expected = SynapseException.class)
```
himasha: Can we check for a specific assert message as well? If not how are we going to differentiate from testCreateTaskManager3() and  testCreateTaskManager4()  as they both throw the same Exception Type?

modules/core/src/test/java/org/apache/synapse/config/xml/MessageProcessorSerializerTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * test serializeMessageProcessor with parameters and assert the OMElement is created.
+     */
+    @Test
+    public void testSerializeMessageProcessor3() {
+        MessageProcessor messageProcessor = new ScheduledMessageForwardingProcessor();
+        messageProcessor.setName("testStore");
+        Map<String, Object> parameters = new HashMap<>();
```
milindaperera: Shall we initialize this parameters variable at the beginning of the function?

modules/core/src/test/java/org/apache/synapse/config/xml/TaskManagerFactoryTest.java
Nirothipan: Fixed this. Thank you

modules/core/src/test/java/org/apache/synapse/config/xml/MessageProcessorSerializerTest.java
Nirothipan: Fixed this.

modules/core/src/test/java/org/apache/synapse/config/xml/TaskManagerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+        String inputXML = "<taskManager provider=\"org.apache.synapse.startup.quartz.QuartzTaskManager\">"
+                + "<parameter xmlns=\"http://ws.apache.org/ns/synapse\"  name =\"test\"> test </parameter>"
+                + "</taskManager>";
+        OMElement element = AXIOMUtil.stringToOM(inputXML);
+        TaskManager taskManager = TaskManagerFactory.createTaskManager(element, null);
+        Assert.assertTrue("TaskManager is not created.", taskManager instanceof QuartzTaskManager);
+    }
+
+    @Rule
+    public ExpectedException expectedEx = ExpectedException.none();
```
nuwanw: can you move the variable initialising to beginning of the class

Nirothipan: fixed this

PR #940, opened 2017-11-09
Add tests for SALSession

ThishaniLucas:


modules/core/src/test/java/org/apache/synapse/endpoints/dispatch/SALSessionsTest.java
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * Test class for SALSession
+ */
+public class SALSessionsTest {
+    /**
+     * Test updating session with session id
+     * @throws Exception
+     */
+    @Test
+    public void testUpdateWithId() throws Exception{
```
indikasampath2000: Method signature should format correctly. There should be space before `{`. Please check in the entire class.

ThishaniLucas: Fixed with https://github.com/wso2/wso2-synapse/pull/1100

PR #943, opened 2017-11-09
Shutdown the thread pool started in PassThroughHttpListener

nuwanw:
This will shutdown the thread pools when the listener stop.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceHandler.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceHandler.java
@@ -600,7 +600,8 @@ public class SourceHandler implements NHttpServerEventHandler {
             if (sourceConfiguration.getWorkerPool() != null) {
                 sourceConfiguration.getWorkerPool().shutdown(1000);
             }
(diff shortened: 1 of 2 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceHandler.java
```diff
@@ -589,7 +589,21 @@ public void commitResponseHideExceptions(
...
+    public void stop() {
+        latencyView.destroy();
+        s2sLatencyView.destroy();
+        threadingView.destroy();
+        try {
+            if (sourceConfiguration.getWorkerPool() != null) {
+                sourceConfiguration.getWorkerPool().shutdown(1000);
+            }
+        } catch (InterruptedException e) {
+            log.warn("Error while shutting down worker thread pool. " + e.getMessage());
```
sdkottegoda: shouldn't this be an error log? Also, don't we need to retry the operation if it was interrupted?

nuwanw: Retry will not be successful. And this exception has to be ignored. That is why warn log is added.

PR #954, opened 2017-11-13
Add test case for PayloadFactoryMediatorFactory class

Bhashinee:


modules/core/src/test/java/org/apache/synapse/config/xml/PayloadFactoryMediatorFactoryTest.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+package org.apache.synapse.config.xml;
+
+import org.apache.axiom.om.OMElement;
+import org.apache.axiom.om.util.AXIOMUtil;
+import org.apache.synapse.Mediator;
+import org.apache.synapse.config.xml.PayloadFactoryMediatorFactory;
+import org.junit.Test;
+
```
abeykoon: Shall we remove the additional lines here in imports?

Bhashinee: Done

PR #957, opened 2017-11-15
Add a unit test case for ProxyServiceMessageReceiver

abeykoon:
## Purpose
Add unit tests for ProxyServiceMessageReceiver

### DIFF
--- a/modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
+++ b/modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
@@ -1,0 +1,116 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied. See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.synapse.core.axis2;
+
+
+import org.apache.axiom.om.OMAbstractFactory;
+import org.apache.axiom.om.OMElement;
+import org.apache.axiom.om.util.AXIOMUtil;
+import org.apache.axiom.util.UIDGenerator;
+import org.apache.axis2.AxisFault;
+import org.apache.axis2.context.ConfigurationContext;
+import org.apache.axis2.context.OperationContext;
+import org.apache.axis2.context.ServiceContext;
+import org.apache.axis2.description.InOutAxisOperation;
+import org.apache.axis2.engine.AxisConfiguration;
+import org.apache.synapse.MessageContext;
+import org.apache.synapse.config.SynapseConfiguration;
+import org.apache.synapse.config.xml.SequenceMediatorFactory;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+
+/**
+ * Unit tests to test methods in class ProxyServiceMessageReceiver
+ */
+public class ProxyServiceMessageReceiverTest {
+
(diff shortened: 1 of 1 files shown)


modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
```diff
@@ -0,0 +1,101 @@
...
+        MessageContext messageContext = createMessageContext();
+        Axis2SynapseEnvironment synapseEnvironment = getMockedSynapseEnvironment();
+
+        //set required values
+        MessageContextCreatorForAxis2.setSynConfig(new SynapseConfiguration());
+        MessageContextCreatorForAxis2.setSynEnv(synapseEnvironment);
+        messageContext.setEnvironment(synapseEnvironment);
+
+        //invoke
+        proxyServiceMessageReceiver.receive(((Axis2MessageContext) messageContext).getAxis2MessageContext());
```
sdkottegoda: Don't we need to assert anything?

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Create a mock SynapseEnvironment object
```
sdkottegoda: Doc comments need to end with a period.

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
```diff
@@ -0,0 +1,101 @@
...
+     * @throws Exception on a message receipt issue
+     */
+    @Test
+    public void testReceive() throws Exception {
+
+        //create ProxyServiceMessageReceiver instance
+        ProxyServiceMessageReceiver proxyServiceMessageReceiver = new ProxyServiceMessageReceiver();
+        proxyServiceMessageReceiver.setProxy(new ProxyService("TestProxy"));
+
+        MessageContext messageContext = createMessageContext();
```
sdkottegoda: Can't use the TestMessageContext for this? Do we need to write a separate method?

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
abeykoon: It is returning nothing. So no need to assert.

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
abeykoon: done.

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
abeykoon: When I do that I am getting
java.lang.ClassCastException: org.apache.synapse.TestMessageContext cannot be cast to org.apache.synapse.core.axis2.Axis2MessageContext

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
sdkottegoda: We shouldn't be asserting only the methods that are returning something. We need to find something to assert. May be a change in a field. Otherwise having the real method implementation empty could also pass the unit test.

abeykoon: Done. Set a property in the inSequence and asserted it once returned from the method.

modules/core/src/test/java/org/apache/synapse/core/axis2/ProxyServiceMessageReceiverTest.java
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @throws Exception on a message receipt issue
+     */
+    @Test
+    public void testReceive() throws Exception {
+
+        //create ProxyServiceMessageReceiver instance
+        ProxyServiceMessageReceiver proxyServiceMessageReceiver = new ProxyServiceMessageReceiver();
+        ProxyService proxyService = new ProxyService("TestProxy");
+        OMElement sequenceAsOM = AXIOMUtil.stringToOM("<inSequence xmlns=\"http://ws.apache.org/ns/synapse\">\n" +
```
sdkottegoda: Line wrapping should happen before an operation

PR #958, opened 2017-11-15
Fix for scheduling inbound task at delete and shutdown

prabushi:
## Purpose
Fix the issue of deleting registry task at cluster node shutdown with hazelcast.shutdownhook.enabled=false.

## Goals
Fixed: https://github.com/wso2/product-ei/issues/1206

## Approach
As the approach, we introduced a new parameter to differentiate shutdown and delete scenarios, and delete the registry task only at inbound endpoint delete. Inbound task at registry will remain at the shutdown scenario.

## Release note
will be released with 6.2.0

## Documentation
https://docs.wso2.com/display/EI611/Clustered+Deployment

## Related PRs
https://github.com/wso2/carbon-mediation/pull/908

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
@@ -26,8 +26,13 @@ import org.apache.axis2.engine.AxisConfiguration;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.synapse.*;
+import org.apache.synapse.ManagedLifecycle;
+import org.apache.synapse.Mediator;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.Startup;
+import org.apache.synapse.SynapseArtifact;
+import org.apache.synapse.SynapseConstants;
+import org.apache.synapse.SynapseException;
 import org.apache.synapse.aspects.flow.statistics.store.CompletedStructureStore;
 import org.apache.synapse.carbonext.TenantInfoConfigProvider;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
```diff
@@ -1447,10 +1452,10 @@ public synchronized void destroy(boolean preserverState) {
         }
 
         //destroy inbound endpoint
-		for (InboundEndpoint endpoint : getInboundEndpoints()) {
-			endpoint.destroy();
-		}         
-        
+        for (InboundEndpoint endpoint : getInboundEndpoints()) {
```
VijithaEkanayake: Shall we explain the purpose of this code segment in a comment?

modules/core/src/main/java/org/apache/synapse/inbound/InboundEndpoint.java
```diff
@@ -140,13 +140,28 @@ private InboundProcessorParams populateParams() {
         return inboundProcessorParams;
     }
 
-    public void destroy() {
+    public void destroy(boolean removeTask) {
```
VijithaEkanayake: Please add doc comment to here?

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
prabushi: Updated

modules/core/src/main/java/org/apache/synapse/inbound/InboundEndpoint.java
prabushi: Updated.

isudana:
Shall we update the description of the PR ?

prabushi:
@isudana Updated the description.

PR #967, opened 2017-11-23
Cache schemas in Validate Mediator

sohaniwso2:
This solves the issue of re-initializing all the schemas for each and every request instead of caching the schemas
Public Jira : https://wso2.org/jira/browse/ESBJAVA-4958


modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
```diff
@@ -220,7 +240,24 @@ public boolean mediate(MessageContext synCtx) {
...
+                         */
+                        if (!cachedJsonSchemaMap.containsKey(cachedJsonSchemaKey.toString())) {
+                            cachedJsonSchemaMap.put(cachedJsonSchemaKey.toString(), cachedJsonSchema);
+                            /*
+                             * Removes the existing cached schema and adds the
+                             * new cached schema This is used when editing a
+                             * registry resource or when the cache expires
+                             */
+                        } else if (cachedJsonSchemaMap.containsKey(cachedJsonSchemaKey.toString())) {
+
```
VijithaEkanayake: unnecessary new line, Shall we remove it?

sohaniwso2: done

PR #971, opened 2017-11-27
Refactor doc comments in Script Mediator

malakaganga:
## Purpose
Refactor doc comments in Script Mediator


modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/CommonScriptMessageContext.java
```diff
@@ -451,6 +503,13 @@ public void removeProperty(String key, String scope) {
 
     }
 
+    /**
+     * Add content type to the headers.
```
isudana: Is this correct ? method name doesn't match with this

PR #973, opened 2017-12-05
Fix NPE in message processor when message body is empty

ThishaniLucas:
When a message without SOAP body is stored in a JDBC store and processed by a message processor, a Null Pointer Exception is thrown at the message converting end.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
@@ -358,7 +358,9 @@ public final class MessageConverter {
             response.getFirstElement().detach();
         }
 
-        soapEnvelope.getBody().addChild(response.getFirstElement().getFirstElement());
+        if (response.getFirstElement().getFirstElement() != null) {
+            soapEnvelope.getBody().addChild(response.getFirstElement().getFirstElement());
+        }
 
         return soapEnvelope;
     }


modules/core/src/main/java/org/apache/synapse/message/store/impl/commons/MessageConverter.java
```diff
@@ -358,7 +358,9 @@ private static SOAPEnvelope getSoapEnvelope(String soapEnvelpe) {
             response.getFirstElement().detach();
         }
 
-        soapEnvelope.getBody().addChild(response.getFirstElement().getFirstElement());
+        if (response.getFirstElement().getFirstElement() != null) {
```
isudana: Can you please double check the logic of the entire method ? It looks to me logic does not differentiate soap headers and body properly.

ThishaniLucas: The header element is processed and detached in the code block above the new addition. Now the response.getFirstElement() returns the body.

isudana: ok, got it

isudana: What will happen if we get a envelope without a body here ?

isudana: though it is mandatory to have a soap body for a soap envelop, in case of someone send without a body we shouldn't through a NPE. 

can that happen in this case ?

ThishaniLucas: No it won't. We're constructing the soap envelop from the default soap envelop of the soapfactory. It has an empty header and an empty body. If the response has no body, then the envelop will be returned with the available empty soap body.

isudana: ok. great.

PR #989, opened 2017-12-29
Fix product-ei/issues/1140 & product-ei/issues/1600

Bhashinee:
Related issues: https://github.com/wso2/product-ei/issues/1140
https://github.com/wso2/product-ei/issues/1600

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceResponse.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceResponse.java
@@ -248,13 +248,14 @@ public class SourceResponse {
             String contentType = headers.get(HTTP.CONTENT_TYPE) != null ?
                     headers.get(HTTP.CONTENT_TYPE).toString() : null;
             // Stream should be preserved to support disable chunking for SOAP based responses. This checks
-            // whether chunking is disabled and response Content-Type is SOAP.
-            boolean shouldPreserve = isChunkingDisabled && isSOAPContentType(contentType);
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceResponse.java
```diff
@@ -248,8 +248,9 @@ private void calculateContentlengthForChunckDisabledResponse(MessageContext resp
             String contentType = headers.get(HTTP.CONTENT_TYPE) != null ?
                     headers.get(HTTP.CONTENT_TYPE).toString() : null;
             // Stream should be preserved to support disable chunking for SOAP based responses. This checks
-            // whether chunking is disabled and response Content-Type is SOAP.
-            boolean shouldPreserve = isChunkingDisabled && isSOAPContentType(contentType);
+            // whether chunking is disabled and response Content-Type is SOAP or FORCE_HTTP_1.0 is true.
+            boolean shouldPreserve =
```
abeykoon: What about a variable name like shouldPreserveChunking? forceChunking?

Bhashinee: In here we are preserving the stream (if chunking is disabled and response Content-Type is SOAP, the stream is preserved).

abeykoon: Then a name like preserveStream? We need to add what to be preserved in the variable name.

Bhashinee: Changed the name into preserveStream.

PR #996, opened 2018-01-04
Fix for ESBJAVA-4443

VijithaEkanayake:
When the synapse.streaming.xpath.enabled is set to true, Synapse always trying to build the message as XML. We only need to build the message as XML if the Content-Type is Streaming Xpath support one. Related JIRA : https://wso2.org/jira/browse/ESBJAVA-4538

## Purpose
Build the request messages in a correct manner when the synapse.streaming.xpath.enabled is set to true. By default Synapse always trying to build the message as XML.

## Approach
Check whether content-type of the request is streaming XPath supported one or not and engage correct builders accordingly.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
@@ -29,6 +29,8 @@ import org.apache.axiom.om.impl.llom.OMDocumentImpl;
 import org.apache.axiom.om.impl.llom.OMElementImpl;
 import org.apache.axiom.om.impl.llom.OMTextImpl;
 import org.apache.axiom.om.impl.llom.util.AXIOMUtil;
+import org.apache.axiom.soap.SOAP11Constants;
+import org.apache.axiom.soap.SOAP12Constants;
 import org.apache.axiom.soap.SOAPEnvelope;
 import org.apache.axiom.soap.impl.dom.factory.DOMSOAPFactory;
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
```diff
@@ -368,7 +372,15 @@ public String stringValueOf(MessageContext synCtx) {
...
+                    if (!isStreamingXpathSupportedContentType(contentType) &&
+                            !Boolean.TRUE.equals(PassThroughConstants.MESSAGE_BUILDER_INVOKED)) {
+                        RelayUtils.buildMessage(axis2MC);
+                    } else {
+                        inputStream = getMessageInputStreamPT(axis2MC);
+                    }
+                } catch (XMLStreamException e) {
+                    handleException("Error occurred while building the message from the message context", e);
                 } catch (IOException e) {
                     log.error("Error occurred while obtaining input stream from the message context", e);
```
milindaperera: Don't we need to handle IOExceptions?

milindaperera: Not related to this fix, but we can improve rest while fixing

VijithaEkanayake: If we handle the exception here, the flow will be stopped and faultHandler will be executed. Even IOException has occurred, flow should be continued because Xpath evaluation will be continued over existing SOAP envelope when the input stream is null.

PR #998, opened 2018-01-07
Refactor ConcurrencyThrottlingUtils class

GDLMadushanka:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/ConcurrencyThrottlingUtils.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/ConcurrencyThrottlingUtils.java
@@ -47,10 +47,10 @@ public class ConcurrencyThrottlingUtils {
         if (isConcurrencyThrottleEnabled != null && isConcurrencyThrottleEnabled) {
             ConcurrentAccessController concurrentAccessController = (ConcurrentAccessController) synCtx
                     .getProperty(SynapseConstants.SYNAPSE_CONCURRENT_ACCESS_CONTROLLER);
+            int available = concurrentAccessController.incrementAndGet();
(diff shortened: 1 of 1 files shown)


Drifftr:
Shall we add the public jira in the commit description?

modules/core/src/main/java/org/apache/synapse/util/ConcurrencyThrottlingUtils.java
```diff
@@ -47,9 +47,9 @@ public static void decrementConcurrencyThrottleAccessController(MessageContext s
         if (isConcurrencyThrottleEnabled != null && isConcurrencyThrottleEnabled) {
             ConcurrentAccessController concurrentAccessController = (ConcurrentAccessController) synCtx
                     .getProperty(SynapseConstants.SYNAPSE_CONCURRENT_ACCESS_CONTROLLER);
+            int available = concurrentAccessController.incrementAndGet();
+            int concurrentLimit = concurrentAccessController.getLimit();
             if (log.isDebugEnabled()) {
-                int available = concurrentAccessController.incrementAndGet();
-                int concurrentLimit = concurrentAccessController.getLimit();
                 log.debug("Concurrency Throttle : Connection returned" + " :: " + +available + " of available of"
```
harsha89: There are two + signs and space is missing at the end of "of available of". Please correct them as well

GDLMadushanka: +1 Done

harsha89: @GDLMadushanka you might have missed adding " of available of" space and make this " of available of " ?

PR #1001, opened 2018-01-08
Fix wso2/product-ei/#1664

malakaganga:
Throws a ScriptException In processJSONPayload method when trying
to parse a JSON payload and it fails, rather than
throwing a SynapseException.

Fixes: https://github.com/wso2/product-ei/issues/1664

to parse a JSON payload and it fails

## Security checks

### DIFF
--- a/modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/ScriptMediator.java
+++ b/modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/ScriptMediator.java
@@ -410,7 +410,8 @@ public class ScriptMediator extends AbstractMediator {
                     jsonObject = this.jsEngine.eval('(' + scriptWithJsonParser + ')');
                 }
             } catch (ScriptException e) {
-                throw new SynapseException("Invalid JSON payload", e);
+                throw new ScriptException("Invalid JSON payload", e.getFileName(), e.getLineNumber(),
+                        e.getColumnNumber());
             }
         } else if (jsonString != null) {
             String jsonPayload = jsonParser.parse(jsonString).toString();


modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/ScriptMediator.java
```diff
@@ -410,7 +410,7 @@ private void processJSONPayload(MessageContext synCtx, ScriptMessageContext scri
                     jsonObject = this.jsEngine.eval('(' + scriptWithJsonParser + ')');
                 }
             } catch (ScriptException e) {
-                throw new SynapseException("Invalid JSON payload", e);
+                throw new ScriptException("Invalid JSON payload");
```
milindaperera: shall we add exception object as follows:
throw new ScriptException("Invalid JSON payload",e);
Otherise upstream will lose stacktrace

bsenduran: By not showing the `e`, is there any useful information get hidden, if so better to keep the e

sdkottegoda: Also, would it be helpful if we log the payload?

malakaganga: Hi,
Since we are throwing ScriptException the constructor (message, e) doesn't exists what we can propagate is the message , filename , linenumber etc.

isudana: IMO, logging the payload is not useful.

malakaganga: With the e it propagates only the line number and column number where the error occurs so in this case the important thing to propagate would be that the payload is a invalid one.

vinok88: Then please include the filename and line number inside the error message.

malakaganga:
Related to: wso2/product-ei/issues/1605

PR #1015, opened 2018-01-17
Fix handling POST request without Content-Type

VijithaEkanayake:
When POST request comes without Content-Type recipient may assume it as application/octet-stream ((rfc2616)). This fixes https://wso2.org/jira/browse/ESBJAVA-2183

## Purpose
Handle POST requests without Content-Type as defined in HTTP spec.

## Approach

When EI receives a POST without content type, EI considers content type as application/x-www-form-urlencoded. As per the spec, EI should handle it as application/octet-stream. The "octet-stream" is used to indicate that a body contains arbitrary binary data. So to handle content-type according to the HTTP 1.1, EI should be shipped enabling the binary relay for octet-stream Content-Type i.e configuring BinaryBuilder and ExpandingFormatter as Builder and Formatter for application/octet-stream in axis2.xml file

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
@@ -183,10 +183,15 @@ public class ServerWorker implements Runnable {
 	 */
 	public SOAPEnvelope handleRESTUrlPost(String contentTypeHdr) throws FactoryConfigurationError {
 	    SOAPEnvelope soapEnvelope = null;
-	    String contentType = contentTypeHdr!=null?TransportUtils.getContentType(contentTypeHdr, msgContext):null;
-	    if (contentType == null || "".equals(contentType) || HTTPConstants.MEDIA_TYPE_X_WWW_FORM.equals(contentType)) {
-	        contentType = contentTypeHdr != null ? contentTypeHdr:HTTPConstants.MEDIA_TYPE_X_WWW_FORM;
-	        msgContext.setTo(new EndpointReference(request.getRequest().getRequestLine().getUri()));
+        String contentType = contentTypeHdr != null ? TransportUtils.getContentType(contentTypeHdr, msgContext) : null;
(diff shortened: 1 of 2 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
```diff
@@ -183,10 +183,15 @@ private boolean isRequestToFetchWSDL() {
...
 	public SOAPEnvelope handleRESTUrlPost(String contentTypeHdr) throws FactoryConfigurationError {
 	    SOAPEnvelope soapEnvelope = null;
-	    String contentType = contentTypeHdr!=null?TransportUtils.getContentType(contentTypeHdr, msgContext):null;
-	    if (contentType == null || "".equals(contentType) || HTTPConstants.MEDIA_TYPE_X_WWW_FORM.equals(contentType)) {
-	        contentType = contentTypeHdr != null ? contentTypeHdr:HTTPConstants.MEDIA_TYPE_X_WWW_FORM;
-	        msgContext.setTo(new EndpointReference(request.getRequest().getRequestLine().getUri()));
+        String contentType = contentTypeHdr != null ? TransportUtils.getContentType(contentTypeHdr, msgContext) : null;
+        //When POST request doesn't contain a Content-Type,
+        // recipient should consider it as application/octet-stream (rfc2616)
+        if (contentType == null || "".equals(contentType)) {
```
wggihan: Better if we can use isEmpty method in here.

chamilaadhi:
We have the **ERROR - ServerWorker Error while building message for REST_URL request** error printing in APIM latest release (as mentioned in #1860 ) and it gets printed for every request without any content type.

wso2/wso2-synapse#1015 has added a fix to apply a content-type for POST request if the content type is missing. but handleRESTUrlPost() gets executed for GET request as well.

Seems like the logic in here https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java#L155 is causing the issue. handleRESTUrlPost() gets called for GET method as well

PR #1025, opened 2018-01-23
Make proxy unsecure when adding policy without enabling security

ThishaniLucas:
Resolves the issues : https://wso2.org/jira/browse/ESBJAVA-4459 and https://github.com/wso2/product-ei/issues/1703

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
@@ -259,52 +259,50 @@ public class ProxyServiceFactory {
             proxy.setResourceMap(ResourceMapFactory.createResourceMap(wsdl));
         }
 
-        Iterator policies = elem.getChildrenWithName(
-                new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "policy"));
-        while (policies.hasNext()) {
-            Object o = policies.next();
-            if (o instanceof OMElement) {
-                OMElement policy = (OMElement) o;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
```diff
@@ -259,52 +259,50 @@ public static ProxyService createProxy(OMElement elem, Properties properties) {
...
-                            pi.setType(PolicyInfo.MESSAGE_TYPE_IN);
-                        } else if ("out".equals(type.getAttributeValue())) {
-                            pi.setType(PolicyInfo.MESSAGE_TYPE_OUT);
-                        } else {
-                            handleException("Undefined policy type for the policy with key : "
-                                    + key.getAttributeValue());
+        if (isSecured(elem) && (getPolicies(elem) != null)) {
+            Iterator policies = getPolicies(elem);
+            while (policies.hasNext()) {
+                Object o = policies.next();
```
chanikag: Shall we have a meaningful name?

ThishaniLucas: Done

lahirus:
"createProxy" method is too long and almost the whole class, which is not a problem with this PR. Better if we can break it meaningfully.

lahirus:
There are no DEBUG logs in the class. Again it is not introduced from this PR, but better to add some DEBUG logs.

ThishaniLucas:
Closing the issue since we need more testing.

PR #1030, opened 2018-01-25
Remove inline $ctx support in payload factory

bsenduran:
https://wso2.org/jira/browse/ESBJAVA-4647

## Purpose
Since the $ctx is a part of synapse xpath, and introducing full xpath
support in payload factory inline will introduce complexity.
Currenty argumet supports fully synapse xpath capability.
Hence removing this partial support.

## Goals
Remove partial support of synapse xpath

## Approach
Remove inline $ctx support in payload factory

## User stories
N/A

## Release note
N/A

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 - Integration tests
  Verified all mediation profile integration tests 

## Security checks
N/A

## Samples
N/A

## Related PRs
https://github.com/wso2/wso2-synapse/pull/261

## Migrations (if applicable)
N/A

## Test environment
N/A
 
## Learning
N/A

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
@@ -82,7 +82,6 @@ public class PayloadFactoryMediator extends AbstractMediator {
 
     private List<Argument> pathArgumentList = new ArrayList<Argument>();
     private Pattern pattern = Pattern.compile("\\$(\\d)+");
-    private Pattern ctxPattern = Pattern.compile("\\$(ctx.[^,\"'<>\n}\\]]*)");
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -293,7 +227,11 @@ private void replace(String format, StringBuffer result, MessageContext synCtx)
         String replacementValue = null;
         Matcher matcher;
 
-        matcher = pattern.matcher(format);
+        if (mediaType != null && (mediaType.equals(JSON_TYPE) || mediaType.equals(TEXT_TYPE))) {
```
erandacr: Can't we change this to 
`JSON_TYPE.equals(mediaType) || TEXT_TYPE.equals(mediaType) `
and avoid the null check?

bsenduran: +1

PR #1037, opened 2018-02-02
Fix product-ei#1815

prabushi:
Validate if the API resource methods are empty at CApp deployment.

## Purpose
Fixed https://github.com/wso2/product-ei/issues/1815


prabushi:
At the management console we validate and does not allow to create an API resource without specifying the methods. So, do we need to allow API resources without any methods?

prabushi:
Will add this validation from the developer studio side.

prabushi:
Added devstudio API resource methods' validation at https://github.com/wso2/devstudio-tooling-esb/pull/317

PR #1055, opened 2018-03-02
Fix securing proxy based on published wsdl security policies

himasha:
Introduced a property to specify whether to preserve/remove security policies
of the published wsdl, so that the proxy can specify its own policies.
Fixes wso2/product-ei#1841

## Purpose
When creating a proxy with publishWSDL option, and if the publish wsdl contains security policies, the proxy is automatically secured with those policies. This does not let the user to configure a different security policy to secure the proxy. 

## Goals
Let users add different secured policies to a proxy where the published wsdl already contains security policies of its own.

## Documentation
https://docs.wso2.com/display/EI611/Publishing+a+WSDL  (Need to update with new attribute introduced)

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
@@ -247,8 +247,7 @@ public class ProxyServiceFactory {
                         log.error(msg);
                         handleException(msg, e);
                     }
-                }
-                 else {
+                } else {
                     OMElement wsdl11 = wsdl.getFirstChildWithName(
                             new QName(WSDLConstants.WSDL1_1_NAMESPACE, "definitions"));
                     if (wsdl11 != null) {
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceFactory.java
```diff
@@ -242,7 +247,8 @@ public static ProxyService createProxy(OMElement elem, Properties properties) {
                         log.error(msg);
                         handleException(msg, e);
                     }
-                } else {
+                }
+                 else {
```
lahirus: Minor formatting issue.

himasha: Fixed.

modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
```diff
@@ -448,6 +465,29 @@ public AxisService buildAxisService(SynapseConfiguration synCfg, AxisConfigurati
         // if a WSDL was found
         if (wsdlElement != null) {
             OMNamespace wsdlNamespace = wsdlElement.getNamespace();
+            // if preservePolicy is set to 'false', remove the security policy content of publish wsdl
```
lahirus: buildAxisService method is too long. Better to at least take the new part to a different methode.

himasha: +1. Moved code to new method.

PR #1058, opened 2018-03-08
Fix issues in concurrent access throttling

milanperera:

### DIFF
--- a/modules/commons/src/main/java/org/apache/synapse/commons/throttle/core/ConcurrentAccessController.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/throttle/core/ConcurrentAccessController.java
@@ -95,8 +95,4 @@ public class ConcurrentAccessController implements Serializable {
 		}
 
 	}
-
-	public int get() {
-		return counter.get();
-	}
 }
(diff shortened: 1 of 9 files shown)


modules/commons/src/main/java/org/apache/synapse/commons/throttle/core/ConcurrentAccessController.java
```diff
@@ -49,28 +49,24 @@ public ConcurrentAccessController(int limit) {
 	 *
 	 * @return the previous value
 	 */
```
lahirus: Indentation is broken in ConcurrentAccessController file.

lahirus:
Could we remove commented off code lines in this PR, even though they are not related?
- //    private final Map<String, AxisCallback> callbackStore;  // will made thread safe in the constructor
- //        callbackStore = Collections.synchronizedMap(new HashMap<String, AxisCallback>());

IMO we don't need the commented code. If we really need them. Fix the indentation similar to other comments.

modules/commons/src/main/java/org/apache/synapse/commons/throttle/core/ConcurrentAccessController.java
milanperera: Fixed in [1]

[1] https://github.com/wso2/wso2-synapse/pull/1058/commits/8f45351b62d0bcb2c1e13c3fc788890e50c32850

PR #1073, opened 2018-03-24
Fix continuation call to avoid getting false incorrectly

nuwanw:
This fix will avoid continuation call getting set to false even when the non blocking call mediators are present in the configuration while redeploying and undeploying car file having blocking call mediators.


nuwanw:
resolve https://github.com/wso2/product-ei/issues/2053

PR #1075, opened 2018-03-27
Avoid the error when sequence template is undeployed

ThishaniLucas:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/deployers/TemplateDeployer.java
+++ b/modules/core/src/main/java/org/apache/synapse/deployers/TemplateDeployer.java
@@ -35,6 +35,7 @@ import org.apache.synapse.mediators.template.TemplateMediator;
 
 import javax.xml.namespace.QName;
 import java.io.File;
+import java.util.Map;
 import java.util.Properties;
 
 public class TemplateDeployer extends AbstractSynapseArtifactDeployer {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/deployers/TemplateDeployer.java
```diff
@@ -214,22 +215,22 @@ public void undeploySynapseArtifact(String artifactName) {
...
-                        log.debug("Undeploying template is not of endpoint type. Undeployer will now check " +
-                                  "for sequence template for the key: " + artifactName);
-                    }
-                } else {
-                    //different error hence stop undeployment/report failure
+            Map<String, Template> templates = getSynapseConfiguration().getEndpointTemplates();
+            if(templates.containsKey(artifactName)){
+                try {
+                    st = getSynapseConfiguration().getEndpointTemplate(artifactName);
+                } catch (SynapseException e) {
```
Manuri: Why are we catching the exception and just throwing it?

modules/core/src/main/java/org/apache/synapse/deployers/TemplateDeployer.java
```diff
@@ -214,22 +215,22 @@ public void undeploySynapseArtifact(String artifactName) {
...
-                        log.debug("Undeploying template is not of endpoint type. Undeployer will now check " +
-                                  "for sequence template for the key: " + artifactName);
-                    }
-                } else {
-                    //different error hence stop undeployment/report failure
+            Map<String, Template> templates = getSynapseConfiguration().getEndpointTemplates();
+            if (templates.containsKey(artifactName)) {
+                try {
+                    st = getSynapseConfiguration().getEndpointTemplate(artifactName);
+                } catch (SynapseException e) {
```
Manuri: Why are we catching the exception and just throwing it?

ThishaniLucas: Actually yes. Because there is nothing we should do except stop the undeployment flow.

PR #1101, opened 2018-05-02
Fix github.com/wso2/product-ei/issues/2149

GDLMadushanka:


riyafa:
Shall we change the commit message as Fix https://github.com/wso2/product-ei/issues/2149

PR #1123, opened 2018-06-07
Updated removeSeqContinuationState method in ContinuationStackManager

tbottelbergs:
For anonymous sequences, the reliant continuation state is removed (as
opposed to popping the continuation state stack) after processing. This is an error,
since no reliant continuation state is added when the anonymous sequence
is started: the method addSeqContinuationState only pushes a
continuation state onto the stack for non-anonymous sequences, but
doesn't do anything for anonymous sequences.

I have created this based on the v2.1.7wso2v15 tag and tested in wso2ei-6.1.1 (which used the v2.1.7wso2v15 synapse-core). Please check this fix, since I was unable to run the synapse-core unit tests on my system.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/continuation/ContinuationStackManager.java
+++ b/modules/core/src/main/java/org/apache/synapse/continuation/ContinuationStackManager.java
@@ -93,9 +93,7 @@ public class ContinuationStackManager {
         if (synCtx.isContinuationEnabled() && !synCtx.getContinuationStateStack().isEmpty()) {
             if (!SequenceType.ANON.equals(seqType)) {
                 ContinuationStackManager.popContinuationStateStack(synCtx);
-            } else {
-                removeReliantContinuationState(synCtx);
-            }
+            } 
         }
     }
 


isudana:
Hi Tim,

Even though reliant continuation state is not added when entering an an anonymous sequence, it get added by the parent of some anonymous sequences. 
For an example if we consider a clone mediator, clone mediator adds a reliant continuation state before entering a Target (which is an anonymous sequence). When mediating through the continuation stack after call mediator invocation, while returning from the anonymous sequence, we need to clear the related reliant continuation state from the stack, as control will not go back to the parent (in this case clone mediator) of the anonymous sequence. 

So current behaviour is correct. 

Thanks.

tbottelbergs:
Hi Isuru,

Thanks for checking. However, could you also look at the Jira issue I created: https://wso2.org/jira/browse/ESBJAVA-5307. In there, I described a situation in which this will cause a problem (and possible infinite loops). We currently have this problem in one of our customer projects.
In this case, we are not using a clone mediator, but a foreach mediator within a filter mediator. What actually happens here, is that the anonymous sequence within the foreach loop removes a child continuation state for every loop, but this child continuation state is never added by the foreach loop. This results in the child continuation state that was added by the filter mediator to be removed by the anonymous sequence within the foreach loop. 
If at some point (e.g. by the file connector within the foreach loop) a new child continuation state is added, the position indication of the filter mediator within the sequence is reset, which causes the sequence to continue at an arbitrary point after the filter mediator.
It seems to me that the responsibility for adding or removing a child continuation state should be kept within the same entity: or adding and removing a child continuation state is done within the sequence mediator, or it is done by the parent (e.g. a clone or foreach mediator), but not a mix of both.

Regards,

Tim

PR #1132, opened 2018-06-25
Fix product-ei#2335

prabushi:
Populate class mediator properties which are defined as values while deploying the carbon application. Properties which are defined as expressions will be initialized at the mediate method.

## Purpose
Fix https://github.com/wso2/product-ei/issues/2335

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/MediatorPropertyFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/MediatorPropertyFactory.java
@@ -23,6 +23,7 @@ import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.Mediator;
 import org.apache.synapse.SynapseException;
 import org.apache.synapse.mediators.MediatorProperty;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/MediatorPropertyFactory.java
```diff
@@ -82,7 +83,11 @@
                     throw new SynapseException(msg);
 
                 } else {
-                    prop.setValue(attValue.getAttributeValue());
+                    String attributeValue = attValue.getAttributeValue();
+                    prop.setValue(attributeValue);
+                    if (mediator != null) {
```
sdkottegoda: Can we avoid this null check? Whether the mediator is null or not is determined at the compile time itself.

prabushi: This mediator parameter is introduced to distinguish the usage of method at the class mediator. This is a common method used for log, xslt, and endpoint factories too. By default, log. xslt, endpoint have mediator parameter as null. Usage of this method will be derived at runtime. So, we need to have this null check.

prabushi:
Related doc issue https://github.com/wso2/product-ei/issues/2504

PR #1134, opened 2018-06-28
Mediation cache clear when validate mediator redeployment

JanakaPushpakumara:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
@@ -46,6 +46,7 @@ import org.apache.synapse.config.SynapseConfigUtils;
 import org.apache.synapse.config.SynapseConfiguration;
 import org.apache.synapse.config.xml.SynapsePath;
 import org.apache.synapse.continuation.ContinuationStackManager;
+import org.apache.synapse.core.SynapseEnvironment;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.mediators.AbstractListMediator;
 import org.apache.synapse.mediators.FlowContinuableMediator;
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
```diff
@@ -563,6 +566,27 @@ public boolean mediate(MessageContext synCtx,
         return result;
     }
 
+    /**
+     * Initialize child mediators recursively
+     *
+     * @param se synapse environment
+     */
+    public void init(SynapseEnvironment se) {
+        if (Boolean.parseBoolean(se.getSynapseConfiguration().getProperty(SYNAPSE_VALIDATE_MEDIATOR_REDEPLOYMENT_CACHE_CLEAR))){
```
nuwanw: shall we add false as default value

malakaganga: Shall we fix the formatting issue?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
```diff
@@ -563,6 +566,27 @@ public boolean mediate(MessageContext synCtx,
...
+            for(Value schemaKey :schemaKeys) {
```
malakaganga: Shall we fix the formatting issue?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
JanakaPushpakumara: It already provides boolean false as the default value

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
JanakaPushpakumara: fixed

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ValidateMediator.java
JanakaPushpakumara: fixed

JanakaPushpakumara:
Fixed issues with the pr https://github.com/wso2/wso2-synapse/pull/1135

Issue #1114, opened 2018-05-31
Overflow blob and connection handling when transferring large files


dkgkim:
**Description:**

If disableChunking is "true", the length is calculated by writing a file to the java temp directory via OverflowBlob for large file transfers.

However, the connection can be made to the `stale` state during the completion of the` OverflowBlob` because the process is performed after the http connection is established through the `deliveryAgent.submit ()` function.

Therefore, there are two solutions.

1. Execute `overflowBlob.writeTo (out)` and then check `pipe.isStale`, and throw an exception to handle the fault in the mediation flow.

2. Call `deliveryAgent.submit ()` after OverflowBlob processing so that the connection does not fall into `stale` state.

We can also apply both options.

This is about [PassThroughHttpSender#L270](https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java#L270) and [PassThroughHttpSender#L372](https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java#L372).

Once the solution to the problem is decided, it seems that we can proceed with the source code revision.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

dkgkim:
fixed

PR #1137, opened 2018-07-11
delay connection while overflow blob process

dkgkim:
**Description:**

This is about Issue https://github.com/wso2/wso2-synapse/issues/1114 (Overflow blob and connection handling when transferring large files).

If disableChunking is "true", the length is calculated by writing a file to the java temp directory via OverflowBlob for large file transfers.

However, the connection can be made to the `stale` state during the completion of the` OverflowBlob` because the process is performed after the http connection is established through the `deliveryAgent.submit ()` function.

Therefore, there are two solutions.

1. Execute `overflowBlob.writeTo (out)` and then check `pipe.isStale`, and throw an exception to handle the fault in the mediation flow.

2. Call `deliveryAgent.submit ()` after OverflowBlob processing so that the connection does not fall into `stale` state.

I applied both options.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
@@ -437,6 +437,8 @@ public class PassThroughHttpSender extends AbstractHandler implements TransportS
                     }
                 }
             }
+        } else {
+            deliveryAgent.submit(msgContext, epr);
         }
     }
 	
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
@@ -269,8 +269,7 @@ public class PassThroughHttpSender extends AbstractHandler implements TransportS
                             (String) msgContext.getProperty(PassThroughConstants.ORGINAL_CONTEN_LENGTH)));
                 }
 
-                deliveryAgent.submit(msgContext, epr);
-                sendRequestContent(msgContext);
+                sendRequestContent(msgContext, epr);
             } else {
                 handleException("Cannot send message to " + AddressingConstants.Final.WSA_NONE_URI);
             }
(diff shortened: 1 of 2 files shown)


isudana:
Hi Kim,

Thanks for the contribution. We will review and merge the pull request.

Thanks.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
```diff
@@ -336,60 +335,77 @@ public void removePreserveHttpHeader(String headerName) {
...
-                        e.printStackTrace();
-                    }
+    private boolean waitForReady(final MessageContext msgContext) {
+        synchronized (msgContext) {
+            while (!Boolean.TRUE.equals(msgContext.getProperty(PassThroughConstants.WAIT_BUILDER_IN_STREAM_COMPLETE)) &&
+                   !Boolean.TRUE.equals(msgContext.getProperty("PASSTHRU_CONNECT_ERROR"))) {
+                try {
+                    msgContext.wait();
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
```
milindaperera: please avoid using printStackTrace(). Let's log the error stack trace with a message using log.error()

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
```diff
@@ -336,60 +335,77 @@ public void removePreserveHttpHeader(String headerName) {
...
+                                        (Boolean) msgContext.getProperty(PassThroughConstants.FORCE_POST_PUT_NOBODY))) {
+                            pipe.setSerializationCompleteWithoutData(true);
+                        } else {
+                            pipe.setSerializationComplete(true);
+                        }
                     }
-
-                } else {
+                } catch (IOException e) {
+                    // TODO Auto-generated catch block
```
milindaperera: please remove autogenerated comments

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
dkgkim: fixed

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
dkgkim: fixed

PR #1139, opened 2018-07-16
Adding the name of the proxy services in error message when the policy reference doesn't exist to help diagnose problems when deploying multiple ProxyService

SLAVEH3R0:
handleException("NoSecurity Policy found from key " + key + " in Axis2 service \\"" + name + "\\"");

>Instead of debugging with this :
>>INFO - DependencyTracker Local entry : gov:/ws-policy/policyname.xml was added to the Synapse configuration successfully
>>[2018-07-13 08:26:42,305] []  WARN - SynapseConfigUtils Cannot convert null to a StreamSource


isudana:
Hi SLAVEH3R0,

Thanks for the contribution. We will review and merge the pull request.

Thanks.

PR #1176, opened 2018-10-01
Fix wso2/product-ei/issues/2705

GDLMadushanka:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/MessageHelper.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/MessageHelper.java
@@ -74,28 +74,34 @@ public class MessageHelper {
 
     private static Log log = LogFactory.getLog(MessageHelper.class);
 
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/synapse/util/MessageHelper.java
```diff
@@ -414,7 +426,9 @@ public static MessageContext cloneMessageContextForAggregateMediator(MessageCont
         }
         // XXX: always this section must come after the above step. ie. after applying Envelope.
         // That is to get the existing headers into the new envelope.
-        JsonUtil.cloneJsonPayload(mc, newMC);
+        if (isCloneJson) {
+            JsonUtil.cloneJsonPayload(mc, newMC);
+        }
```
abeykoon: If isCloneJson=false we clone the context, but just place the reference to the original json playload without cloning? What if somebody used payload factory inside iterate mediator? Then on next iteration it is applied on the replaced payload?

PR #1177, opened 2018-10-05
fix for https://github.com/wso2/product-apim/issues/3595

isharac:


modules/core/src/main/java/org/apache/synapse/rest/Resource.java
```diff
@@ -21,11 +21,9 @@
 import org.apache.axiom.util.UIDGenerator;
 import org.apache.axis2.Constants;
 import org.apache.http.HttpHeaders;
+import org.apache.http.HttpStatus;
 import org.apache.http.protocol.HTTP;
-import org.apache.synapse.ManagedLifecycle;
-import org.apache.synapse.Mediator;
-import org.apache.synapse.MessageContext;
-import org.apache.synapse.SynapseException;
+import org.apache.synapse.*;
```
isudana: Import only required classes. Importing everything (.*) is not a best practice.

Issue #1197, opened 2018-11-06
 JSON to XML transformation with payload factory does not if emojis are contains in the JSON Payload


LahiruLS:
**Description:**

When the payload factory mediator is used to transform from JSON to XML and if the JSON payload contains emojis an exception is thrown. 
When replacing XML special characters to related escape sequences, Unicode characters greater than 0x7f are currently escaped to their numerical value. In this case, Emojis are converted to numerical values(decimal value of them). 

**Suggested Labels:**

XML
Special Characters
Escape Characters

**Affected Product Version:**
EI 6.1.1
synapse-core-2.1.7-wso2v15

**OS, DB, other environment details and versions:**
The issue was reproducible in Ubuntu 18.04 LST    

**Steps to reproduce:**

1. Start EI 6.1.1 and log into management console.
2. Create a synapse artifact with a payload factory and transform JSON payload to XML payload.
3. For example below sample api,
```
<api xmlns="http://ws.apache.org/ns/synapse" name="Sample" context="/someurl/test">
   <resource methods="POST">
      <inSequence>
         <property name="test" expression="json-eval(test.info)" scope="default" type="STRING"/>
         <property name="ContentType" value="text/xml" scope="axis2" type="STRING"/>
         <payloadFactory media-type="xml">
            <format>
               <dat:testTag xmlns:dat="http://ws.wso2.org/dataservice">
                  <dat:InnerTag>$1</dat:InnerTag>
               </dat:testTag>
            </format>
            <args>
               <arg evaluator="xml" expression="get-property('test')"/>
            </args>
         </payloadFactory>
         <respond/>
      </inSequence>
      <outSequence>
         <log level="full"/>
      </outSequence>
      <faultSequence/>
   </resource>
</api>
```
    4. Use the follow payload to invoke the api,
```
{"test":
	{
		"info":"😂"
	}
}
```

LahiruLS:
This occurs due to the Apache commons-lang version at EI 6.1.1 is 2.4 and the escape XML special characters method is not properly implemented in the particular commons-lang version. In the java doc of the method(escapeXml) [1], it stated that the Unicode characters greater than 0x7f are currently escaped to their numerical \\u equivalent and this may change in future releases. So Emojis are in that particular range which is replaced with their decimal values.

[1] - [](https://github.com/apache/commons-lang/blob/aa9727e15966bac899aa18733ef85c4aded19f0c/src/java/org/apache/commons/lang/StringEscapeUtils.java#L577)

PR #1201, opened 2018-11-11
Fix wso2/wso2-synapse#1197

Drifftr:
## Purpose
Use commons-text to escape xml messages in Payload factory. Previously XML envelop is escaped through commons-lang which didn't support emojis. commons-text support escaping special chars in both XML v1.0 and XML v1.1.
Issue - wso2/wso2-synapse#1197

## Goals
Support Emoji chars in payloads.

## Approach
Used commons-text v1.6 instead of commons-lang v2.4 which supports emojis and escaping other special characters.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
@@ -31,7 +31,6 @@ import org.apache.axiom.soap.SOAPEnvelope;
 import org.apache.axis2.AxisFault;
 import org.apache.axis2.Constants;
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -660,4 +664,49 @@ public boolean isContentAltering() {
...
+     */
+    private String escapeXMLEnvelope(MessageContext msgCtx, String value) {
+        String xmlVersion = "1.0"; //Default is set to 1.0
+
+        try {
+            xmlVersion = checkXMLVersion(msgCtx);
+        } catch (IOException e) {
+            log.error("Error reading message envelope", e);
+        } catch (SAXException e) {
+            log.error("Error parsing message envelope", e);
```
milindaperera: We could catch both exceptions together

milindaperera: Didn't notice messages are different. Sorry for noise

a5anka:
@Drifftr Shall we merge this?

PR #1202, opened 2018-11-12
Add fix for issue in setting content-type 

vithu30:
## Purpose
This PR contains changes to solve issues in adding content-type for the requests that are sent without content-type header. Current implementation adds content-type as application/octet-stream for all incoming requests that do not have content-type header, except GET method. This fix introduces the changes to add content-type as application/octet-stream for the POST request which has body but not content-type.
Relevant specification can be found in [here](https://tools.ietf.org/html/rfc2616#page-43). It fixes the issue - [Content-Type:application/octet-stream is added when no Content-type header is provided #3926](https://github.com/wso2/product-apim/issues/3926)

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
@@ -189,10 +189,20 @@ public class ServerWorker implements Runnable {
         // When POST request doesn't contain a Content-Type,
         // recipient should consider it as application/octet-stream (rfc2616)
         if (contentType == null || contentType.isEmpty()) {
-            contentType = PassThroughConstants.APPLICATION_OCTET_STREAM;
+            String httpMethod = (String) msgContext.getProperty(HTTP_METHOD);
             // Temp fix for https://github.com/wso2/product-ei/issues/2001
-            if (HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(HTTP_METHOD))) {
+            // Else if block is added to prevent the content-type assignment as application/octet-stream
+            // for all incoming requests. Current fix sets content-type as application/octet-stream only
+            // for POST requests which has body and no content-type
+            // as per spec: https://tools.ietf.org/html/rfc2616#page-43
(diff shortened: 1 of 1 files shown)


a5anka:
@vithu30 Please try to use short commit titles in the future. 

https://chris.beams.io/posts/git-commit/

a5anka:
@vithu30 Did we add a test case for this?

vithu30:
Closing the PR since the fix is reverted in the latest update

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
```diff
@@ -189,10 +190,19 @@ public SOAPEnvelope handleRESTUrlPost(String contentTypeHdr) throws FactoryConfi
...
-            if (HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(HTTP_METHOD))) {
+            // Else if block is added to prevent the content-type assignment as application/octet-stream
+            // for all incoming requests. Current fix sets content-type as application/octet-stream only
+            // for POST requests which has body and no content-type
+            // as per spec: https://tools.ietf.org/html/rfc2616#page-43
+            if (HTTPConstants.HTTP_METHOD_GET.equals(httpMethod)) {
                 contentType = HTTPConstants.MEDIA_TYPE_X_WWW_FORM;
+            } else if (HTTPConstants.HTTP_METHOD_POST.equals(httpMethod)) {
+                if (request.isEntityEnclosing() &&
+                        ((HttpEntityEnclosingRequest) request.getRequest()).getEntity().getContentLength() != 0) {
```
hasuniea: What would happen getEntity returns null?

vithu30: isEntityEnclosing() returns whether the request has body or not - [1]. When getEntity() returns null, request does not have body, hence if condition won't get executed
[1] https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceRequest.java#L49

PR #1205, opened 2018-11-16
Add native JSON support for Aggregate mediator

GDLMadushanka:


a5anka:
@GDLMadushanka Shall we close this since @sachithKay is working on an improvement?

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
```diff
@@ -142,4 +144,86 @@ public String getJsonPathExpression() {
...
+     * value, if the path is definite. if the path is not a definite list will
+     * contain multiple element. NULL will return if the path is invalid. Empty
+     * list will return if the path points to null.
+     */
+    private List listValueOf(final InputStream jsonStream) {
+        if (jsonStream == null) {
+            return null;
+        }
+        List result = new ArrayList();
+        Object object;
```
milindaperera: let's define the object variable inside try-catch

GDLMadushanka: +1 fixed

PR #1206, opened 2018-11-19
Resolve wso2/product-ei#2861

rosensilva:
Error sequence is not reached when clone mediator is used in sequential
mode. Please refer https://wso2.org/jira/browse/ESBJAVA-4913

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/eip/Target.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/eip/Target.java
@@ -107,7 +107,7 @@ public class Target {
                 if (log.isDebugEnabled()) {
                     log.debug("Synchronously mediating using the in-lined anonymous sequence");
                 }
-                returnValue = sequence.mediate(synCtx);
+                returnValue = mediateMessage(sequence, synCtx);
             }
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/eip/Target.java
```diff
@@ -270,4 +249,31 @@ public void setStatisticIdForMediators(ArtifactHolder holder){
         }
         StatisticIdentityGenerator.reportingEndBranchingEvent(holder);
     }
+
+    private boolean mediateMessage(SequenceMediator sequenceMediator, MessageContext synCtx) {
+        try {
+            return sequenceMediator.mediate(synCtx);
+        } catch (SynapseException syne) {
+            if (!synCtx.getFaultStack().isEmpty()) {
+                log.warn("Executing fault handler due to exception encountered");
```
a5anka: Isn't it better to add the exception stack traces?

rosensilva: fixed with https://github.com/wso2/wso2-synapse/pull/1215

PR #1209, opened 2018-11-30
Add endpoint resolver

shyabithd:
Description:
As the Micro integrator executable we are going a provide a docker image and since the image is self contained, we must pack artifacts into the image itself as well. Current implementation requires endpoints to be specified in the synapse configuration but in cloud we cannot specify endpoint uris before hand since we are not aware about the host or the port of the endpoint service.
Solution:
Instead of specifying uri in the synapse config, the proposal is to use variable in places where uri is not known before hand.

Fixes: wso2/product-ei/issues/2735


modules/core/src/test/java/org/apache/synapse/endpoints/resolvers/ResolverFactoryTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.synapse.config.xml.endpoints.resolvers.SystemResolver;
+import org.junit.contrib.java.lang.system.EnvironmentVariables;
+import org.junit.runner.RunWith;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+
+@RunWith(PowerMockRunner.class)
+public class ResolverFactoryTest extends TestCase {
+
+    public void testGetSystemResolver() {
```
abeykoon: Shall we add a description what each test does?

modules/core/src/test/java/org/apache/synapse/endpoints/resolvers/ResolverFactoryTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public void testGetDefaultResolver() {
+        String synapseVariable = "https://localhost:9443/services";
+        Resolver resolver = ResolverFactory.getInstance().getResolver(synapseVariable);
+        assertNotNull(resolver);
+        assertTrue(resolver instanceof DefaultResolver);
+        assertEquals(synapseVariable, resolver.resolve());
+    }
+
+    public void testGetUnknownResolver() {
```
ThishaniLucas: Shall we use expected exception here, instead of asserting the exception?

shyabithd: done

PR #1210, opened 2018-12-04
Fixed the 'http DELETE is not working without its content-type header'

HiranyaKavishani:
## Purpose
These fixes are resolved the issue in delete service since it is not working without the content-type header.

## Approach
when payload stream of DeleteRequest is empty, it needs to return false in the method.

#https://github.com/wso2/product-apim/issues/4033

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
@@ -220,20 +220,55 @@ public class RelayUtils {
      * @param msgContext MessageContext
      * @return whether the request is a DELETE without payload
      */
-    public static boolean isDeleteRequestWithoutPayload (MessageContext msgContext) {
+    public static boolean isDeleteRequestWithoutPayload (MessageContext msgContext) throws AxisFault {
         if (PassThroughConstants.HTTP_DELETE.equals(msgContext.getProperty(Constants.Configuration.HTTP_METHOD))) {
 
             //If message builder not invoked (Passthrough may contain entity body) OR delete with payload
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
```diff
@@ -236,6 +238,39 @@ public static boolean isDeleteRequestWithoutPayload (MessageContext msgContext)
...
+
+        boolean isEmpty = false;
+        BufferedInputStream bufferedInputStream = (BufferedInputStream) messageContext
+                .getProperty(PassThroughConstants.BUFFERED_INPUT_STREAM);
+        if (bufferedInputStream != null) {
+            try {
+                bufferedInputStream.reset();
+                bufferedInputStream.mark(0);
+            } catch (Exception e) {
+                // just ignore the error
```
mushthaq33: Why are we ignoring the exception here?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
```diff
@@ -236,6 +237,39 @@ public static boolean isDeleteRequestWithoutPayload (MessageContext msgContext)
...
+            }
+
+        } else {
+            final Pipe pipe = (Pipe) messageContext.getProperty(PassThroughConstants.PASS_THROUGH_PIPE);
+            if (pipe != null) {
+                InputStream in = pipe.getInputStream();
+                bufferedInputStream = new BufferedInputStream(in);
+                // TODO: need to handle properly for the moment lets use around 100k
+                // buffer.
```
nalakas: shall we put one line comment here?

HiranyaKavishani: This change cannot be found in the current file

PR #1212, opened 2018-12-04
Revert "Fixed the 'http DELETE is not working without its content-type header'"

malinthaprasan:
Reverts wso2/wso2-synapse#1210

Reverted due to the regression: The outgoing payload is getting dropped for the DELETE requests with payload and content type.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
@@ -149,8 +149,8 @@ public class RelayUtils {
             try {
                 bufferedInputStream.reset();
                 bufferedInputStream.mark(0);
-            } catch (IOException e) {
-                handleException("Error while checking bufferedInputStream", e);
+            } catch (Exception e) {
+                // just ignore the error
             }
 
         } else {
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
```diff
@@ -149,8 +149,8 @@ public static void builldMessage(MessageContext messageContext, boolean earlyBui
             try {
                 bufferedInputStream.reset();
                 bufferedInputStream.mark(0);
-            } catch (IOException e) {
-                handleException("Error while checking bufferedInputStream", e);
+            } catch (Exception e) {
+                // just ignore the error
```
a5anka: Why are we ignoring the exception here? Won't this make debugging issues harder?

a5anka: @malinthaprasan Shall we fix it using a separate commit eventhough this comes from the revert?

indikasampath2000: +1. Let's do it as a different PR. But I guess we have to merge it as it introduces a regression.

PR #1215, opened 2018-12-11
Add stack trace to the error log

rosensilva:


modules/core/src/main/java/org/apache/synapse/mediators/eip/Target.java
```diff
@@ -255,7 +255,7 @@ private boolean mediateMessage(SequenceMediator sequenceMediator, MessageContext
             return sequenceMediator.mediate(synCtx);
         } catch (SynapseException syne) {
             if (!synCtx.getFaultStack().isEmpty()) {
-                log.warn("Executing fault handler due to exception encountered");
+                log.warn("Executing fault handler due to exception encountered" + syne);
```
abeykoon: Shall we modify as log.warn("Executing fault handler due to exception encountered" ,  syne); ?

rosensilva: fixed

Issue #1226, opened 2018-12-17
Missing keystore configured in multi-ssl-profile.xml breaking all the API calls


nimanthag:
**Description:**
During a recent incident API Cloud, we came across a situation where a missing Keystore for a profile configured in multi-ssl-profile.xml breaking the all calls through the gateway. As a solution, we need early detection system to handle this kind of scenarios. For that if we encounter such error we need to immediately stop the server startup. If it is a dynamic update we have to proceed the server work ignoring that dynamic change.

**Affected Product Version:**
APIM 2.5.0

**OS, DB, other environment details and versions:**    
MAC
**Steps to reproduce:**
1) Configure Mutual SSL according to https://docs.wso2.com/display/AM210/Mutual+SSL+Support+for+API+Gateway link.
2) Change the jks file name in <APIM_HOME>/repository/deployment/server/multi_ssl_profiles.xml. 
3) Start the server. You will see file not found exception, but server starts ignoring that.

PR #1227, opened 2018-12-17
Fix invalid configuration in multi-ssl-profile.xml breaking all the API Call issue

nimanthag:
### Purpose
Resolves https://github.com/wso2/wso2-synapse/issues/1226
### Goal
When there is a configuration error in the initial startup. The server will shut down immediately.
If there is a configuration error in dynamically updated SSL, the server will continue to process ignoring that update.
### Approach
Throw runtime exception if there is an issue in the configuration. This will stop the server startup. If there is an error in dynamic change path, catch that specific runtime exception and ignore the update and proceed.


npamudika:
This PR fixes https://github.com/wso2/product-ei/issues/3584

PR #1245, opened 2019-01-21
Changed JmsConsumer receive wait time

Ibaqu:
- Changed the consumer.receive(1) to consumer.receive(500) for the
getMessage function defined in MessageProcessorAdminService to work

## Purpose
To make the PoisonMessage service functionality work


modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java
```diff
@@ -83,7 +83,7 @@ public MessageContext receive() {
         }
 
         try {
-            Message message = consumer.receive(1);
+            Message message = consumer.receive(500);
```
a5anka: Do you think 500 is enough for an average environment?

PR #1249, opened 2019-01-30
 Fix issue with auto lock release

tharinduwijewardane:
Update acquire-lock logic to consider transport.vfs.AutoLockRelease and transport.vfs.AutoLockReleaseInterval parameters and release the lock as expected.
Fixes wso2/product-ei#3153

Refactor VFSUtils

### DIFF
--- a/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
@@ -58,7 +58,11 @@ public class VFSUtils {
     private static final Log log = LogFactory.getLog(VFSUtils.class);
 
     private static final String STR_SPLITER = ":";
-    
+
+    private static final String LOCK_FILE_SUFFIX = ".lock";
+
+    private static final String FAIL_FILE_SUFFIX = ".fail";
+
     /**
      * URL pattern
      */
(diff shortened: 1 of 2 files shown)


modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
```diff
@@ -270,7 +244,60 @@ public synchronized static boolean acquireLock(FileSystemManager fsManager, File
...
+    private static boolean createLockFile(byte[] lockValue, FileObject lockObject, String fullPath)
+            throws FileSystemException {
+        // write a lock file before starting of the processing, to ensure that the
+        // item is not processed by any other parties
+        lockObject.createFile();
+        OutputStream stream = lockObject.getContent().getOutputStream();
+        try {
+            stream.write(lockValue);
+            stream.flush();
+            stream.close();
```
Asitha: Shall we close the stream on finally block? Otherwise, this can lead to a resource leak when an exception occurring on write() or flush method

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -553,6 +554,7 @@ private void scanFileOrDirectory(final PollTableEntry entry, String fileURI) {
                         		Thread.sleep(iFileProcessingInterval);
                         	}catch(InterruptedException ie){
```
Asitha: shall we fix the code formatting?

modules/commons/src/main/java/org/apache/synapse/commons/vfs/VFSUtils.java
tharinduwijewardane: fixed

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
tharinduwijewardane: fixed

PR #1251, opened 2019-01-31
Fix wso2/product-ei/issues/3163

malakaganga:
## Purpose

Fixes: https://github.com/wso2/product-ei/issues/3163


modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
```diff
@@ -662,4 +664,32 @@ private void registerFaultHandler(org.apache.synapse.MessageContext synCtx) {
...
     }
+    /**
+     * Invoke Synapse Handlers
+     *
+     * @param synCtx synapse message context
+     * @return whether flow should continue further
+     */
+    private boolean invokeHandlers(org.apache.synapse.MessageContext synCtx) {
+        Iterator<SynapseHandler> iterator = synCtx.getEnvironment().getSynapseHandlers().iterator();
+        if (iterator.hasNext()) {
```
milindaperera: Shall we add debug log when executing each handler?, that will improve UX when troubleshooting.
Add logging in success flow as well

PR #1268, opened 2019-03-07
Add native json support for Iterate mediator

arunans23:
## Purpose
Adding native json support for Iterate mediator.
https://github.com/wso2/product-ei/issues/3311

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
@@ -19,6 +19,10 @@
 
 package org.apache.synapse.mediators.eip.splitter;
 
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParser;
+import com.jayway.jsonpath.JsonPath;
 import org.apache.axiom.om.OMElement;
 import org.apache.axiom.om.OMNode;
 import org.apache.axiom.soap.SOAPEnvelope;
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/IterateMediator.java
```diff
@@ -117,70 +129,143 @@ public boolean mediate(MessageContext synCtx) {
...
         try {
-            // get a copy of the message for the processing, if the continueParent is set to true
-            // this original message can go in further mediations and hence we should not change
-            // the original message context
-            SOAPEnvelope envelope = MessageHelper.cloneSOAPEnvelope(synCtx.getEnvelope());
 
-            synCtx.setProperty(id != null ? EIPConstants.EIP_SHARED_DATA_HOLDER + "." + id :
-                               EIPConstants.EIP_SHARED_DATA_HOLDER, new SharedDataHolder());
+            // check whether expression contains jsonpath or xpath and process according to it
+            if (expression != null && expression instanceof SynapseJsonPath) {
```
sdkottegoda: Redundant null check

arunans23: This was added to prevent null expressions getting inside the else block. If we remove null check, null expressions will get executed inside the else block.

sdkottegoda: Still, a null expression would go inside the else block, right? This will evauluate to true only if the expression is not null.

PR #1273, opened 2019-03-14
Revert proxy addition in axis2 service startup failure

sdkottegoda:
## Purpose
When a proxy service is deployed with a faulty security policy reference, the proxy service is correctly initialized and added to the synapse configuration. The build then fails when building the axis2 service.

The proxy service is marked as failed and not removed from the synapse configuration. When redeploying the same service, it is picked up as already deployed, but the proxy service doesn't work.

The fix is to remove the proxy service from the synapse configuration if the axis2 service fails to startup.
Resolves: https://github.com/wso2/product-ei/issues/2525
Related to: https://github.com/wso2/carbon-mediation/pull/1125

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
+++ b/modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
@@ -24,6 +24,7 @@ import org.apache.axis2.deployment.DeploymentException;
 import org.apache.axis2.description.AxisService;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.SynapseException;
 import org.apache.synapse.transport.customlogsetter.CustomLogSetter;
 import org.apache.synapse.config.xml.MultiXMLConfigurationBuilder;
 import org.apache.synapse.config.xml.ProxyServiceFactory;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
```diff
@@ -79,32 +80,35 @@ public String deploySynapseArtifact(OMElement artifactConfig, String filePath,
...
 
+                    if (!proxy.isStartOnLoad() || !axisService.isActive()) {
+                        proxy.stop(getSynapseConfiguration());
+                        log.info("ProxyService named '" + proxy.getName()
+                                 + "' has been stopped as startOnLoad parameter is set to false");
+                    }
+                    return proxy.getName();
+                } catch (SynapseException e) {
+                    getSynapseConfiguration().removeProxyService(proxy.getName());
+                    throw e;
```
Nirothipan: Shall we use handleSynapseArtifactDeploymentError method here ?

Nirothipan: Caught in https://github.com/wso2/wso2-synapse/pull/1273/files#diff-771ac85f9b9e01d365ddef0c512dfdeaR124

PR #1277, opened 2019-03-21
Fix inability to add comments inside mediators in synapse

dinuish94:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/BeanMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/BeanMediatorFactory.java
@@ -74,6 +74,8 @@ public class BeanMediatorFactory extends AbstractMediatorFactory {
             handleException("'action' attribute of Bean mediator is required.");
         }
 
+        addAllCommentChildrenToList(elem, mediator.getCommentsList());
+
         return mediator;
     }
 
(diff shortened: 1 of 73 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/BeanMediatorFactory.java
```diff
@@ -74,6 +74,8 @@ public Mediator createSpecificMediator(OMElement elem, Properties properties) {
             handleException("'action' attribute of Bean mediator is required.");
         }
 
+        addAllCommentChildrenToList(elem, mediator.getCommentsList());
+
```
chanikag: Remove these new lines

modules/core/src/main/java/org/apache/synapse/config/xml/SequenceMediatorFactory.java
```diff
@@ -69,6 +69,7 @@ public SequenceMediator createAnonymousSequence(OMElement elem, Properties prope
         }
         addChildren(elem, seqMediator, properties);
         seqMediator.setSequenceType(SequenceType.ANON);
+
```
chanikag: Unnecessary new line?

modules/core/src/main/java/org/apache/synapse/config/xml/BeanMediatorFactory.java
dinuish94: Fixed in https://github.com/wso2/wso2-synapse/pull/1439

modules/core/src/main/java/org/apache/synapse/config/xml/SequenceMediatorFactory.java
dinuish94: Fixed in https://github.com/wso2/wso2-synapse/pull/1439

dinuish94: Fixed in https://github.com/wso2/wso2-synapse/pull/1439

PR #1298, opened 2019-04-03
Add comments to Proxy Services and APIs

dinuish94:
## Goals
Introduces the ability to persist comments within 'proxy' and 'api' tags.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
@@ -258,7 +258,7 @@ public class ProxyService implements AspectConfigurable, SynapseArtifact {
     private AxisService axisService;
 
     /**
-     * Comment Texts List associated with the proxy service
+     * Holds the list of comments associated with the proxy service.
      */
     private List<String> commentsList = new ArrayList<String>();
 
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
@@ -258,6 +258,11 @@ public class ProxyService implements AspectConfigurable, SynapseArtifact {
     private AxisService axisService;
 
     /**
+     * Comment Texts List associated with the proxy service
+     */
+    private List<String> commentsList = new ArrayList<String>();
+
+    /**
      * Constructor
      *
      * @param name the name of the Proxy service
@@ -1425,4 +1430,12 @@ public class ProxyService implements AspectConfigurable, SynapseArtifact {
 		}
 		StatisticIdentityGenerator.reportingEndEvent(proxyId, ComponentType.PROXYSERVICE, holder);
 	}
+
+    public List<String> getCommentsList() {
+        return commentsList;
+    }
+
+    public void setCommentsList(List<String> commentsList) {
+        this.commentsList = commentsList;
+    }
 }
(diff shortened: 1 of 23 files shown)


modules/core/src/main/java/org/apache/synapse/util/CommentListUtil.java
```diff
@@ -0,0 +1,45 @@
+package org.apache.synapse.util;
```
madushadhanushka: Need license header

modules/core/src/main/java/org/apache/synapse/util/CommentListUtil.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.axiom.om.*;
```
madushadhanushka: Shall we import only required classes

modules/core/src/main/java/org/apache/synapse/core/axis2/ProxyService.java
```diff
@@ -257,6 +257,11 @@ public void setPreservePolicy(String preservePolicy) {
 
     private AxisService axisService;
 
+    /**
+     * Comment Texts List associated with the proxy service
+     */
```
isudana: Please refer to following to learn conventions in doc comments.
https://www.oracle.com/technetwork/java/javase/tech/index-137868.html

modules/core/src/main/java/org/apache/synapse/util/CommentListUtil.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class CommentListUtil {
+
+    /**
+     * Adds comments in Synapse to a list of comments
+     *
+     * @param el    OMElement containing the comments
+     * @param commentList List to be which the comments should be added to
+     */
+    public static void addAllCommentChildrenToList(OMElement el, List<String> commentList) {
```
isudana: I think we need a better method name here. Something like "populateComments"

modules/core/src/main/java/org/apache/synapse/endpoints/AbstractEndpoint.java
```diff
@@ -51,12 +51,7 @@
 import org.apache.synapse.transport.passthru.util.RelayConstants;
 import org.apache.synapse.transport.passthru.util.RelayUtils;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.Stack;
+import java.util.*;
```
Nirothipan: Please don't use wild card imports

modules/core/src/main/java/org/apache/synapse/endpoints/Template.java
```diff
@@ -51,6 +51,11 @@
 
     private boolean isEdited;
 
+    /**
+     * Holds the list of comments associated with the proxy service.
+     */
+    private List<String> commentsList = new ArrayList<String>();
```
sdkottegoda: Can we define this in the class SynapseArtifact, so that every artifacts class couls inherit it?

dinuish94: There are certain classes such as SynapseConfiguration, Entry, ScheduledMessageForwardingProcessor, StartUpController etc that we don't require to persist the comments in. Therefore, is it necessary to add it to this layer?

sdkottegoda: Since this is only being used in classes, it's ok to have the code in the child classes.

PR #1301, opened 2019-04-04
Make CallMediator content aware in blocking mode

Vathsan:
Issues [#1805](https://github.com/wso2/product-ei/issues/1805) and [#3015](https://github.com/wso2/product-ei/issues/3015) can be overcome by adding a log full. Issue [#1805](https://github.com/wso2/product-ei/issues/1805) occurs because the json stream is not set in the MessageContext before the call mediator in blocking mode tries to call the JsonFormatter. Because we set the JSON stream in MessageContext only when building the message. Same as this, in issue [#3015](https://github.com/wso2/product-ei/issues/3015) the xml payload is set in the MessageContext only when building the message.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
@@ -339,9 +339,13 @@ public class CallMediator extends AbstractMediator implements ManagedLifecycle {
         }
     }
 
+    /**
+     * Setting return blocking makes CallMediator access the message’s content in blocking mode when mediating messages
+     * Fixes product-ei #1805, #3015
+     */
     @Override
     public boolean isContentAware() {
-        return false;
+        return blocking;
     }
 
     public boolean isBlocking() {


sdkottegoda:
Shall we explain why we are doing this?

riyafa:
Shall we  "Use the imperative mood in the subject line" ? We can say "Make CallMediator content aware in blocking mode"

malakaganga:
Shall we add the issue ID in the commit message?
Also please add meaningful commit message explaining how and why this is solved.

Vathsan:
> Shall we explain why we are doing this?
Issues [#1805](https://github.com/wso2/product-ei/issues/1805) and [#3015](https://github.com/wso2/product-ei/issues/3015) can be overcome by adding a log full. Issue [#1805](https://github.com/wso2/product-ei/issues/1805) occurs because the json stream is not set in the MessageContext before the call mediator in blocking mode tries to call the JsonFormatter. Because we set the JSON stream in MessageContext only when building the message. Same as this, in issue [#3015](https://github.com/wso2/product-ei/issues/3015) the xml payload is set in the MessageContext only when building the message.
Reference:
https://wso2.markmail.org/message/oos6kvrflz4tobgb?q=list:org%2Ewso2%2Edev

modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
```diff
@@ -341,7 +341,7 @@ public void destroy() {
 
     @Override
     public boolean isContentAware() {
-        return false;
+        return blocking;
```
sdkottegoda: Shall we add a comment explaining the change?

Vathsan: Done

PR #1328, opened 2019-04-24
Fixes Json message being logged as XML when log mediator used.

Nirothipan:
Fixes https://github.com/wso2/product-ei/issues/3775.

In Earlier versions of ESB, it was logged as json and after EI 6.2.0 only this behavior was changed.

This has happened due to the fix https://github.com/wso2/wso2-synapse/pull/768/files which was intended to fix https://wso2.org/jira/browse/ESBJAVA-5202.

This PR basically reverts the fix https://github.com/wso2/wso2-synapse/pull/768/files and fixes the original issue mentioned in  https://wso2.org/jira/browse/ESBJAVA-5202. 

The issue in  https://wso2.org/jira/browse/ESBJAVA-5202. has happened since we have not removed the original json stream from context and after iterating we weren't overriding it as a solution for it this PR removes the json stream from the original context in foreach mediator since we don't support native json in for each.

### DIFF
--- a/modules/commons/src/main/java/org/apache/synapse/commons/json/JsonUtil.java
+++ b/modules/commons/src/main/java/org/apache/synapse/commons/json/JsonUtil.java
@@ -62,7 +62,7 @@ import java.util.Properties;
 public final class JsonUtil {
     private static Log logger = LogFactory.getLog(JsonUtil.class.getName());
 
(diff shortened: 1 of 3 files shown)


Nirothipan:
Please don't merge

PR #1332, opened 2019-04-26
Change system properties and set default port for unit testing framework

sajithaliyanage:
## Purpose
Changed all the environment variables to system properties. Therefore no need to export the env variables in script files(bat and sh). If there is a system property called **synapseTest** , then synapse unit test mode will be enabled. Then it checks the **synapseTestPort** property. If there is a receiving port, the framework validates it and start on top of that port.  Else, it sets the port as 9008 (default) and starts executing. 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
@@ -227,27 +227,13 @@ public class Axis2SynapseEnvironment implements SynapseEnvironment {
      * unit test message context and environment initializes
      */
     private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
-        if (Boolean.parseBoolean(System.getenv("UNIT_TEST"))) {
+        if (Boolean.parseBoolean(System.getProperty("synapseTest"))) {
             setUnitTestEnabled(true);
             contextInformation.setServerUnitTestModeEnabled(true);
-            log.info("Synapse unit testing enabled");
+            log.info("Synapse unit testing server enabled");
 
-            //get unit test received port
-            String requestPort = System.getenv("UNIT_TEST_PORT");
-
-            //set default port for TCP server connection
-            int unitTestingAgentPort = 9007;
-            if (requestPort != null && !requestPort.isEmpty()) {
-                try {
-                    unitTestingAgentPort = Integer.parseInt(requestPort);
-                } catch (NumberFormatException e) {
-                    log.error("Given port is not in valid format", e);
-                }
-            }
-
-            //execute synapse unit testing executor
+            //starting UnitTestingExecutor
             UnitTestingExecutor testExecutor = UnitTestingExecutor.getExecuteInstance();
-            testExecutor.setServerPort(unitTestingAgentPort);
             testExecutor.setSynapseConfiguration(this.synapseConfig);
             testExecutor.start();
         }
--- a/modules/core/src/main/java/org/apache/synapse/unittest/TCPServer.java
+++ b/modules/core/src/main/java/org/apache/synapse/unittest/TCPServer.java
@@ -35,17 +35,30 @@ public class TCPServer {
 
     /**
      * Initializing TCP server for main unit testing server.
-     *
-     * @param port port server starts
      */
-    public void initialize(int port) {
+    public void initialize() {
+        //check unit test received port, if its null then set default port as 9008
+        String requestPort = System.getProperty("synapseTestPort");
+        if (requestPort == null || requestPort.isEmpty()) {
+            requestPort = "9008";
+        }
+
+        //check port for TCP server connection and initialized socket
         try {
-            serverSocket = new ServerSocket(port);
-            log.info("Synapse unit testing agent has been established on port " + port);
-            log.info("Waiting for client request");
+            int unitTestingAgentPort = Integer.parseInt(requestPort);
+            serverSocket = new ServerSocket(unitTestingAgentPort);
+            log.info("Synapse unit testing agent has been established on port " + unitTestingAgentPort);
+            if (log.isDebugEnabled()) {
+                log.debug("Waiting for client request");
+            }
+
+            //allow for the client requests
             acceptConnection();
+        } catch (NumberFormatException e) {
+            log.error("Given TCP port \"" + requestPort + "\" is not in valid format, " +
+                    "failed to start unit testing framework", e);
         } catch (IOException e) {
-            log.error(e);
+            log.error("Error in initializing TCP connection in given port " + requestPort, e);
         }
     }
 
--- a/modules/core/src/main/java/org/apache/synapse/unittest/UnitTestingExecutor.java
+++ b/modules/core/src/main/java/org/apache/synapse/unittest/UnitTestingExecutor.java
@@ -30,7 +30,6 @@ public class UnitTestingExecutor extends Thread {
     private static Logger log = Logger.getLogger(UnitTestingExecutor.class.getName());
     private SynapseConfiguration synapseConfiguration;
     private static UnitTestingExecutor initializeThread = new UnitTestingExecutor();
-    private int executingPort;
 
     /**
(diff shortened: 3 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -235,21 +236,30 @@ private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
...
                     unitTestingAgentPort = Integer.parseInt(requestPort);
-                } catch (NumberFormatException e) {
-                    log.error("Given port is not in valid format", e);
+                } else {
+                    throw new IOException("Starting TCP port can not be empty or null");
                 }
+
+            } catch (NumberFormatException e) {
+                log.error("Given port is not in valid format", e);
+            } catch (IOException e) {
```
Nirothipan: Do we need to catch this exception and continue the logic . We don't have anything to continue thereafter isn't it.

sajithaliyanage: Changed the try block and moved the condition into the try block

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -235,21 +236,27 @@ private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
...
                 }
-            }
 
-            //execute synapse unit testing executor
-            UnitTestingExecutor testExecutor = UnitTestingExecutor.getExecuteInstance();
-            testExecutor.setServerPort(unitTestingAgentPort);
-            testExecutor.setSynapseConfiguration(this.synapseConfig);
-            testExecutor.start();
+            } catch (NumberFormatException e) {
+                log.error("Given port is not in valid format", e);
```
a5anka: Shall we print what tried to parse?

Nirothipan: Also better if we could append the error with "failed to start unit testing framework" since it won't start thereafter.

sajithaliyanage: Done

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -235,21 +236,28 @@ private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
...
-                } catch (NumberFormatException e) {
-                    log.error("Given port is not in valid format", e);
+
+                    if (unitTestingAgentPort > 1023) {
+                        UnitTestingExecutor testExecutor = UnitTestingExecutor.getExecuteInstance();
+                        testExecutor.setServerPort(unitTestingAgentPort);
+                        testExecutor.setSynapseConfiguration(this.synapseConfig);
+                        testExecutor.start();
+                    } else {
+                        log.error("Input a TCP port higher than 1023, failed to start unit testing framework");
```
a5anka: Shall we print the given value here too?

sajithaliyanage: Done

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -235,21 +235,29 @@ private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
...
-                } catch (NumberFormatException e) {
-                    log.error("Given port is not in valid format", e);
+
+                    if (unitTestingAgentPort > 1023) {
+                        UnitTestingExecutor testExecutor = UnitTestingExecutor.getExecuteInstance();
+                        testExecutor.setServerPort(unitTestingAgentPort);
+                        testExecutor.setSynapseConfiguration(this.synapseConfig);
+                        testExecutor.start();
+                    } else {
+                        log.error("Input TCP port \"" + unitTestingAgentPort + "\" must be higher than 1023, " +
```
a5anka: Number 1023 in this line and line 244 seems to be same. Therefore it is better to create a constant and use it in the two places.

sajithaliyanage: Done

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -235,21 +237,29 @@ private void setSeverUnitTestMode(ServerContextInformation contextInformation) {
             //get unit test received port
             String requestPort = System.getenv("UNIT_TEST_PORT");
```
sdkottegoda: Should this be an environment variable? Or a system property? What happens if we decide to run 2 instances in the same machine?

sdkottegoda:
Why do we remove the default port setting? Shall we explain it in the PR description?

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
a5anka: IMO 1e need to have it as a system property for the docker images to work properly.

sajithaliyanage: Changed these to system properties.

sajithaliyanage:
> Why do we remove the default port setting? Shall we explain it in the PR description?

Entire PR description and title changed according to changes.

modules/core/src/main/java/org/apache/synapse/unittest/TCPServer.java
```diff
@@ -36,16 +36,23 @@
...
-    public void initialize(int port) {
+    public void initialize(String requestPort) {
+        //check port for TCP server connection and initialized socket
         try {
-            serverSocket = new ServerSocket(port);
-            log.info("Synapse unit testing agent has been established on port " + port);
+            int unitTestingAgentPort = Integer.parseInt(requestPort);
+            serverSocket = new ServerSocket(unitTestingAgentPort);
+            log.info("Synapse unit testing agent has been established on port " + unitTestingAgentPort);
             log.info("Waiting for client request");
```
sdkottegoda: I think this should be a debug log.

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
```diff
@@ -227,27 +227,20 @@ public void setSeverDebugMode(ServerContextInformation contextInformation) {
...
-            if (requestPort != null && !requestPort.isEmpty()) {
-                try {
-                    unitTestingAgentPort = Integer.parseInt(requestPort);
-                } catch (NumberFormatException e) {
-                    log.error("Given port is not in valid format", e);
-                }
+            log.info("Synapse unit testing server enabled");
+
+            //check unit test received port, if its null then set default port as 9008
+            String requestPort = System.getProperty("synapseTestPort");
```
sdkottegoda: I think we can move this inside UnitTestingExecutor, WDYT?

modules/core/src/main/java/org/apache/synapse/unittest/TCPServer.java
sajithaliyanage: Changed as a debug log

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2SynapseEnvironment.java
sajithaliyanage: Yes, we can remove this from here. Changed it as suggested.

PR #1339, opened 2019-05-01
Fix null pointer issue with commentList

prabushi:
Fix null pointer issue with commentList while creating an endpoint template via devstudio.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceSerializer.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceSerializer.java
@@ -239,7 +239,10 @@ public class ProxyServiceSerializer {
             parent.addChild(proxy);
         }
 
-        CommentListUtil.serializeComments(proxy, service.getCommentsList());
+        List<String> commentsList = service.getCommentsList();
+        if (commentsList != null) {
+            CommentListUtil.serializeComments(proxy, commentsList);
+        }
         return proxy;
     }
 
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/synapse/util/CommentListUtil.java
```diff
@@ -58,10 +58,12 @@ public static void populateComments(OMElement el, List<String> commentList) {
      */
     public static void serializeComments(OMElement parent, List<String> commentList) {
         OMFactory fac = OMAbstractFactory.getOMFactory();
-        for (String comment : commentList) {
-            OMComment commendNode = fac.createOMComment(parent, "comment");
-            commendNode.setValue(comment);
-            parent.addChild(commendNode);
+        if (commentList != null) {
```
a5anka: Why do we get commentList as null here? Shouldn't we pass an empty list when we do not have any items.

prabushi: Updated.

modules/core/src/main/java/org/apache/synapse/config/xml/ProxyServiceSerializer.java
```diff
@@ -239,7 +239,10 @@ public static OMElement serializeProxy(OMElement parent, ProxyService service) {
             parent.addChild(proxy);
         }
 
-        CommentListUtil.serializeComments(proxy, service.getCommentsList());
+        List<String> commentsList = service.getCommentsList();
```
a5anka: In ProxyService the  commentsList is initialized to an empty ArrayList. Therefore can  `service.getCommentsList()` return a null?

PR #1340, opened 2019-05-02
Make Api implement SynapseArtifact

sdkottegoda:
## Purpose
Makes Api implement SynapseArtifact so that it has a description.
Resolves: https://github.com/wso2/product-ei/issues/2727


modules/core/src/main/java/org/apache/synapse/rest/API.java
```diff
@@ -50,14 +51,19 @@
...
+public class API extends AbstractRESTProcessor implements ManagedLifecycle, AspectConfigurable, SynapseArtifact {
 
     private String host;
     private int port = -1;
     private String context;
     private Map<String,Resource> resources = new LinkedHashMap<String,Resource>();
     private List<Handler> handlers = new ArrayList<Handler>();
 
+    /**
+     * The proxy service description. This could be optional informative text about the service
```
isudana: This should be API description

PR #1342, opened 2019-05-03
fix default content-type not working issue

dushansilva:
fix for https://github.com/wso2/product-apim/issues/4611

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
@@ -24,6 +24,7 @@ import org.apache.axis2.AxisFault;
 import org.apache.axis2.Constants;
 import org.apache.axis2.addressing.AddressingConstants;
 import org.apache.axis2.addressing.AddressingHelper;
+import org.apache.axis2.context.ConfigurationContext;
 import org.apache.axis2.context.MessageContext;
 import org.apache.axis2.description.WSDL2Constants;
 import org.apache.axis2.engine.AxisConfiguration;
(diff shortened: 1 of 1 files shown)


sdkottegoda:
Did we run the integration tests in the mediation profile of product-ei against the changes introduced?

dushansilva:
Yes the tests were run and they were passing

sdkottegoda:
> Yes the tests were run and they were passing

:+1:

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/RelayUtils.java
```diff
@@ -247,24 +248,26 @@ public static boolean isDeleteRequestWithoutPayload (MessageContext msgContext)
...
         if (headers.size() != 0 && (headers.get("Content-Type") != null || headers.get("content-type") != null)) {
             contentTypeInRequest = true;
         }
-        
-        // If builder is not invoked, which means the passthrough scenario, we should overwrite the content-type 
+        boolean isDefaultContentTypeEnabled = false;
+        ConfigurationContext configurationContex = msgContext.getConfigurationContext();
+        if (configurationContex != null && configurationContex.getAxisConfiguration()
+                .getParameter(NhttpConstants.REQUEST_CONTENT_TYPE) != null) {
+            isDefaultContentTypeEnabled = true;
```
Nirothipan: What will be the behavior if the defaultContentType is **empty/content** ?

dushansilva: It will still use that content type. Is there some other expected behavior for it?

Nirothipan: Will it be written to the output request with this fix

dushansilva: this fix only checks whether that property is added and if its added it will return true,  setting the content type is already done here 

https://github.com/wso2-support/wso2-synapse/blob/support-2.1.7-wso2v65/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/TargetRequestFactory.java#L115

Nirothipan: > this fix only checks whether that property is added and if its added it will return true, setting the content type is already done here

https://github.com/wso2-support/wso2-synapse/blob/support-2.1.7-wso2v65/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/TargetRequestFactory.java#L115

yes true. but based on this only result only we over write the header.

> It will still use that content type. Is there some other expected behavior for it?

If is is empty/content , we shouldn't overwrite the header with that value and we need to skip that in both response and request. Can you please verify whether it is getting skipped there ?

Also better if we check REQUEST_CONTENT_TYPE is not empty in addition to null check.

dushansilva: if it is empty/content it does not override the header in this case, we only check if the header is null. This is done in the request path.

PR #1348, opened 2019-05-23
Fix [EI-4073] null-pointer in header mediator

abeykoon:
## Purpose

Fixes: wso2/product-ei#4073

## Approach

set an empty string as the header when evaluated value for it via an expression becomes `null` and print a `WARN` message. 

## Automation tests

Not needed as this is a fix for a null-pointer 

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/HeaderMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/HeaderMediator.java
@@ -55,6 +55,7 @@ public class HeaderMediator extends AbstractMediator {
 
     public static final int ACTION_SET = 0;
     public static final int ACTION_REMOVE = 1;
+    private static final String EMPTY_STRING = "";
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/transform/HeaderMediator.java
```diff
@@ -171,7 +177,11 @@ public boolean mediate(MessageContext synCtx) {
         } else if (XMLConfigConstants.SCOPE_TRANSPORT.equals(scope)) {        	
         	String headerName = qName.getLocalPart();
             if (action == ACTION_SET) {
-            	
+                if (value == null) {
+                    value = EMPTY_STRING;
+                    log.warn("Setting HTTP header : " + headerName + " to empty as evaluated value is null");
+                }
```
a5anka: Can we avoid the duplicate code?

PR #1352, opened 2019-05-27
Define a method to map query parameters to message properties

SanojPunchihewa:
## Purpose
Since we need to map the query parameters to message properties in both [InternalApiDispatcher](https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/InternalAPIDispatcher.java) and [Resource](https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/rest/Resource.java), define a common method in [RESTUtils](https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java)

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
+++ b/modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
@@ -99,7 +99,7 @@ public class RESTUtils {
      */
     public static void populateQueryParamsToMessageContext(MessageContext synCtx) {
 
-        String path = RESTUtils.getFullRequestPath(synCtx);
+        String path = getFullRequestPath(synCtx);
         String method = (String) synCtx.getProperty(RESTConstants.REST_METHOD);
 
         int queryIndex = path.indexOf('?');
--- a/modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
+++ b/modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
@@ -21,7 +21,9 @@ package org.apache.synapse.rest;
 import org.apache.axis2.Constants;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpStatus;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -88,6 +92,48 @@ public static String getFullRequestPath(MessageContext synCtx) {
         return url;
     }
 
+    public static void mapQueryParamsToMessageProperties(MessageContext synCtx) {
```
Nirothipan: Shall we rename this as **populateQueryParamsToMessageContext** , ( that will describe more what this method intends to do , WDYT ?

Also shall we add a simple method level comment.

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -88,6 +92,48 @@ public static String getFullRequestPath(MessageContext synCtx) {
...
+
+        String path = RESTUtils.getFullRequestPath(synCtx);
+        String method = (String) synCtx.getProperty(RESTConstants.REST_METHOD);
+
+        int queryIndex = path.indexOf('?');
+        if (queryIndex != -1) {
+            String query = path.substring(queryIndex + 1);
+            String[] entries = query.split(RESTConstants.QUERY_PARAM_DELIMITER);
+            String name = null, value ;
```
Nirothipan: Shall we declare value in separate line, just for readability.

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -88,6 +92,54 @@ public static String getFullRequestPath(MessageContext synCtx) {
...
     }
 
+    /**
+     * Populate Message context properties for the query parameters extracted from the url
+     *
+     * @param synCtx MessageContext of the request
+     */
+    public static void populateQueryParamsToMessageContext(MessageContext synCtx) {
+
+        String path = RESTUtils.getFullRequestPath(synCtx);
```
Nirothipan: We can just call the method here.

  ```suggestion
        String path = RESTUtils.getFullRequestPath(synCtx);
```

PR #1365, opened 2019-06-10
Add CORS support for APIs

milindaperera:
This fix provides CORS support for APIs for as a part of adding  swagger editor support for APIs
Fix: https://github.com/wso2/product-ei/issues/4113


modules/core/src/main/java/org/apache/synapse/rest/Resource.java
```diff
@@ -310,10 +311,17 @@ void process(MessageContext synCtx) {
                             statisticReportingIndex, true);
                 }
                 return;
+            } else {
+                // TODO Need to finalize
```
isudana: what is remaining here to be finalized?

milindaperera: Had to finalize how to handle cors headers when respond mediator is used.

milindaperera:
Closing the PR since we have to do a few changes after reviewing, will reopen once changes are made

PR #1366, opened 2019-06-11
Fix the git issue 4127

nuwanw:
This PR will fix the error that occur when json-eval expression is used in sequence template.

wso2/product-ei#4127

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/template/TemplateContext.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/template/TemplateContext.java
@@ -25,6 +25,7 @@ import org.apache.axiom.om.OMText;
 import org.apache.synapse.MessageContext;
 import org.apache.synapse.mediators.Value;
 import org.apache.synapse.mediators.eip.EIPUtils;
+import org.apache.synapse.util.xpath.SynapseJsonPath;
 
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/template/TemplateContext.java
```diff
@@ -92,7 +93,11 @@ private Object getEvaluatedParamValue(MessageContext synCtx, String parameter, V
                 	}
                     return expression.getExpression();
                 } else {
-                    return resolveExpressionValue(synCtx, expression);
+                    if (expression.getExpression() instanceof SynapseJsonPath) {
```
sdkottegoda: I think overriding resolveObject in SynapseJsonPath will be a cleaner fix.

nuwanw: If we override selectSingleNode() in SynapseJsonPath, then it has to return a String value in selectSingleNode. But method name describe itself as a SingleNode. So we need to refactor the method names as well. 
+1 for refactoring whole path of evaluating the parameters in template. Then it will be more clear.

isudana:
I am merging this PR since this has been already merged to some of the braches. Let's keep the issue https://github.com/wso2/product-ei/issues/4127 open to tracking the progress of refactoring template parameter evaluation logic.

PR #1377, opened 2019-07-08
Implement native JSON support for Foreach mediator

arunans23:
## Purpose

Related to wso2/product-ei#4121

This PR will implement native JSON support for Foreach mediator. This is done with regard to completing native JSON support for all the relevant mediators.

Reference : Architecture email "Implementing native JSON support for Synapse mediators"


milindaperera:
Let's create a doc issue with improvements for doc changes

arunans23:
> Let's create a doc issue with improvements for doc changes

@milindaperera added an issue to track the documentation changes https://github.com/wso2/product-ei/issues/4367

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
```diff
@@ -96,85 +105,140 @@ public boolean mediate(MessageContext synCtx) {
...
-                                " = " + originalEnvelope +
-                                "Initialized foreach counter property, " + counterPropName +
-                                " = " + msgCounter);
+                    " = " + originalEnvelope +
+                    "Initialized foreach counter property, " + counterPropName +
+                    " = " + msgCounter);
         }
 
-        SOAPEnvelope processingEnvelope = synCtx.getEnvelope();
+        if (expression != null && expression instanceof SynapseJsonPath) {
```
sdkottegoda: We can remove this unnecessary null check. instance of returns false if the expression is null.

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
```diff
@@ -96,85 +105,140 @@ public boolean mediate(MessageContext synCtx) {
...
-                                    " messages for processing in sequentially, in a general loop");
+                synLog.traceOrDebug("Splitting with JSONPath : " + expression + " resulted in " +
+                        iterableJsonArray.size() + " elements.");
             }
+            JsonArray modifiedPayloadArray = new JsonArray();
+            for (JsonElement element : iterableJsonArray) {
+                try {
+                    updateIteratedMessage(synCtx, element);
+                } catch (AxisFault axisFault) {
+                    handleException("Error creating an iterated copy of the message", axisFault, synCtx);
```
sdkottegoda: Would it be better if the error message was changed to "Error updating the stream with iterater element" + element details ?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
```diff
@@ -96,85 +105,140 @@ public boolean mediate(MessageContext synCtx) {
...
+                }
+                boolean mediateResult = mediateSequence(synCtx);
```
sdkottegoda: I think the code would be more readable if we move the code from here to line:156 inside the try block. WDYT?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
```diff
@@ -96,85 +105,140 @@ public boolean mediate(MessageContext synCtx) {
...
+                modifiedPayloadArray.add(EIPUtils.tryParseJsonString(parser,
```
sdkottegoda: Shouldn't we be peforming the steps here onwards only if mediateResult == true?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
arunans23: This was added to prevent null expressions getting inside the else block. If we remove null check, null expressions will get executed inside the else block.

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
arunans23: We need this part to execute for both true and false cases. We are incrementing the message counter and adding back the element to the envelope. After that we are breaking the for loop.

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
arunans23: Fixed in https://github.com/wso2/wso2-synapse/pull/1445

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
arunans23: Fixed in https://github.com/wso2/wso2-synapse/pull/1445

modules/core/src/main/java/org/apache/synapse/mediators/builtin/ForEachMediator.java
sdkottegoda: Still, a null expression would go inside the else block, right? This will evauluate to true only if the expression is not null.

PR #1383, opened 2019-07-30
Clearer Error Message logs for JSON Schema Validation

Ibaqu:
## Purpose
To give a more detailed error message when using <property expression="get-property('ERROR_DETAIL')">

## Approach
Made changes to the way errors are set for the ERROR_DETAIL property in the message context


Ibaqu:
Related issue : https://github.com/wso2/product-ei/issues/4433

PR #1386, opened 2019-08-01
Swagger fix for product EI

milindaperera:
## Purpose
Add CORS support for APIs

This fix provide CORS support for APIs for as a part of adding  swagger editor support for APIs
Fix: wso2/product-ei#4113

Add publishSwagger support for API

Fix: wso2/product-ei#3328
This fix will add ability to publish swagger definition for synapse API

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/RespondMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/RespondMediator.java
@@ -2,10 +2,12 @@ package org.apache.synapse.mediators.builtin;
 
 
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseLog;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.core.axis2.Axis2Sender;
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/RespondMediator.java
```diff
@@ -28,6 +31,14 @@ public boolean mediate(MessageContext synCtx) {
 
         synCtx.setTo(null);
         synCtx.setResponse(true);
+
+        // if this is not a response from a proxy service
+        String proxyName = (String) synCtx.getProperty(SynapseConstants.PROXY_SERVICE);
+        if (proxyName == null || "".equals(proxyName)) {
```
dulanjalidilmi: Shall we use isEmpty() method to check empty string?

sameeragunarathne: can we use StringUtils.isEmpty() here?

milindaperera: Fixed, thanks

milindaperera: Fixed, thanks

PR #1397, opened 2019-09-16
Fix issues in correlation logs

msm1992:
## Purpose
This PR will solve the following issues found in the observability feature

NPE when utilizing transports other than HTTP/S
Correlation Id reuse fix wso2/product-apim#4690
REQUEST_HEAD  time issue
BE_LATENCY url typo issue
This will also add an improvement for correlation logs in order to print the Time taken to switch from the IO thread to the passthrough thread

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetContext.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetContext.java
@@ -248,7 +248,7 @@ public class TargetContext {
     }
 
     public static Boolean isCorrelationIdAvailable(NHttpConnection connection) {
-        if (connection != null && connection.getContext() != null && connection.getContext().getAttribute(PassThroughConstants.CORRELATION_ID) != null) {
+        if (connection.getContext().getAttribute(PassThroughConstants.CORRELATION_ID) != null) {
             return true;
         }
         return false;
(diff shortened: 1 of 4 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetContext.java
```diff
@@ -247,4 +246,11 @@ public long updateLastStateUpdatedTime() {
         this.lastStateUpdatedTime = System.currentTimeMillis();
         return this.lastStateUpdatedTime;
     }
+
+    public static Boolean isCorrelationIdAvailable(NHttpConnection connection) {
```
sdkottegoda: Can the connection or the context be null at this point?

msm1992: fixed with https://github.com/wso2/wso2-synapse/pull/1397/commits/4ecdd6e044a7436c57bf20df03ecba59bf9d23aa

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetHandler.java
```diff
@@ -51,6 +51,7 @@
 import java.io.IOException;
 import java.util.Properties;
 
+import static org.apache.synapse.transport.passthru.TargetContext.CONNECTION_INFORMATION;
```
sameeragunarathne: It's better if we can remove static imports and use TargetContext.CONNECTION_INFORMATION in the code to get more readability.

msm1992: fixed with https://github.com/wso2/wso2-synapse/pull/1397/commits/32dd70c5631d28b5d28abc529d5e0bba9d4de498

PR #1401, opened 2019-09-17
Move CORS functionalities into a separate class

sachithKay:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/rest/cors/CORSHelper.java
+++ b/modules/core/src/main/java/org/apache/synapse/rest/cors/CORSHelper.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+        String allowedOriginListStr =
+                SynapsePropertiesLoader.getPropertyValue(RESTConstants.CORS_CONFIGURATION_ACCESS_CTL_ALLOW_ORIGIN, "");
+        allowedOrigins = new HashSet<>(Arrays.asList(allowedOriginListStr.split(",")));
+
+        //Retrieve allowed headers
+        allowedHeaders =
+                SynapsePropertiesLoader.getPropertyValue(RESTConstants.CORS_CONFIGURATION_ACCESS_CTL_ALLOW_HEADERS, "");
+    }
+
+    public static SynapseCORSConfiguration getCORSConfig() {
```
milindaperera: Let's rename this function as getInstance as common practice

modules/core/src/main/java/org/apache/synapse/rest/cors/CORSHelper.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.MessageContext;
+import org.apache.synapse.core.axis2.Axis2MessageContext;
+import org.apache.synapse.rest.RESTConstants;
+
+import java.util.Map;
+import java.util.Set;
+
+public class CORSHelper {
```
milindaperera: Let's add class description

modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
sachithKay: fixed

modules/core/src/main/java/org/apache/synapse/rest/cors/CORSHelper.java
sachithKay: fixed

PR #1446, opened 2019-11-19
Consume and discard the pipe when axisfault occurs in ServerWorker

arunans23:
## Purpose
When we send a request without SoapAction, we do not get the expected message when we have a large payload.

This is because, the pipe is not properly consumed and discarded when a AxisFault occurs in ServerWorker.

Fixes https://github.com/wso2/product-ei/issues/4778


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
```diff
@@ -367,6 +368,7 @@ public void processNonEntityEnclosingRESTHandler(SOAPEnvelope soapEnvelope, Mess
                 AxisEngine.receive(msgContext);
             }
         } catch (AxisFault axisFault) {
+            consumeInputOnAxisFault(msgContext);
```
isudana: let's move this consumeInputOnAxisFault call to handleException() method so that it will cover all the paths that this can happen.

arunans23: Fixed

PR #1455, opened 2019-11-26
Fixed latency calculation

AathmanT:
## Purpose
Fixes: https://github.com/wso2/product-ei/issues/4808


isudana:
This change is completely wrong..! 

What we used to have is correct and it gives the latency added by the ESB excluding the backend. 
latency = (tResDeparture - tReqArrival) - backendLatency;

If you make it, 
(tResDeparture - tReqArrival) + backendLatency;

you are doubling the backendLatency as (tResDeparture - tReqArrival) already contains backend latency.

PR #1461, opened 2019-12-03
Fixing access Logs not rotating when not having locale English

CrowleyRajapakse:
## Issue

product-apim:wso2/product-apim#7029

## Methodology
- Passing locale.ENGLISH when creating SimpleDateFormat object.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
@@ -35,6 +35,7 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Locale;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
```diff
@@ -251,7 +252,7 @@ protected synchronized void open() {
                             Matcher matcher = pattern.matcher(line);
                             if (matcher.find()) {
                                 String date = matcher.group(1);
-                                SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT_STRING);
+                                SimpleDateFormat dateFormat = new SimpleDateFormat(DATE_FORMAT_STRING,Locale.ENGLISH);
```
chamilaadhi: foratting issue

CrowleyRajapakse: fixed by #1496

PR #1467, opened 2019-12-06
Fix resolve localhost IP and add to allowed origins

milindaperera:
Issue: https://github.com/wso2/product-ei/issues/4865

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
+++ b/modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
@@ -18,10 +18,18 @@
 
 package org.apache.synapse.rest.cors;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.SynapseException;
 import org.apache.synapse.config.SynapsePropertiesLoader;
 import org.apache.synapse.rest.RESTConstants;
 
-import java.util.Arrays;
+import java.net.InterfaceAddress;
+import java.net.MalformedURLException;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.URL;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
```diff
@@ -42,7 +50,30 @@ private SynapseCORSConfiguration() {
...
         String allowedOriginListStr =
                 SynapsePropertiesLoader.getPropertyValue(RESTConstants.CORS_CONFIGURATION_ACCESS_CTL_ALLOW_ORIGIN, "");
-        allowedOrigins = new HashSet<>(Arrays.asList(allowedOriginListStr.split(",")));
+        String[] originList = allowedOriginListStr.split(",");
+        for (String origin : originList) {
+            String trimmedOrigin = origin.trim();
+            allowedOrigins.add(trimmedOrigin);
+            try {
+                URL url = new URL(trimmedOrigin);
+                if (url.getHost().equals("localhost")) {
```
sameeragunarathne: shall we interchange the constant first in the if check to avoid NPEs

modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
```diff
@@ -42,7 +50,30 @@ private SynapseCORSConfiguration() {
...
+                        NetworkInterface nInterface = networkInterfaces.nextElement();
+                        for (InterfaceAddress iAddr : nInterface.getInterfaceAddresses()) {
+                            URL localUrl =
+                                    new URL(url.getProtocol(), iAddr.getAddress().getHostAddress(), url.getPort(), "");
+                            allowedOrigins.add(localUrl.toString());
+                        }
+                    }
+                }
+            } catch (MalformedURLException e) {
+                LOG.error("Provided origin URL "+ trimmedOrigin + " is malformed", e);
```
sameeragunarathne: only logging is enough here?

modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
milindaperera: NPE never happens since we create new instance just before the if block.

modules/core/src/main/java/org/apache/synapse/rest/cors/SynapseCORSConfiguration.java
milindaperera: Fixed, I was under impression upstream code won't handle the exception. With further investigation found it will execute the fault sequence. So we can throw instead of logging

PR #1478, opened 2020-01-21
Implement graceful shutdown for passthrough transport

arunans23:
## Purpose
Currently Passthrough transport does not support Graceful Shutdown. (ie Accepted requests will not be served properly if the shutdown is triggered.)

Resolves https://github.com/wso2/product-ei/issues/4717

## Approach
There are some metrics collected in Passthrough Transport when request responses are handled. A new metric is introduced in the PassThroughTransportMetricsCollector which will keep count of the current active requests that are being served by the server. When the shutdown is triggered, the server will stop accepting new requests and will wait until the accepted requests are served.

Graceful shutdown explanation (1)

- When a request is accepted, it will increment the metric inside _requestReceived_ method.
- When a request with a body comes, and if an error occurs when reading the stream, the metric will be decremented inside _inputReady_ method.
- When a no entity response is sent, the metric will get decremented in _responseReady_ method.
- When a response with entity is sent, the metric will get decremented in _outputReady_ method.
- When a connection timeout occurs from client to ESB server, the metric will get decremented in _timeout_ method. 

With the implementation of this feature, the server might take sometime to shutdown if there are any active requests that are not served. The server will shutdown **with a worse case time of a value that is equal to the Socket Timeout** that is defined in the passthru-http.properties.

If anyone wish to disable this feature due to development purposes, they may start the server with a System property : **-DgracefulShutdown=false**. The server will always have graceful shutdown enabled by default.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpListener.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpListener.java
@@ -59,6 +59,7 @@ import org.apache.http.HttpHost;
 import org.apache.http.impl.nio.reactor.DefaultListeningIOReactor;
 import org.apache.http.nio.reactor.IOReactorException;
 import org.apache.http.nio.reactor.IOReactorExceptionHandler;
+import org.apache.http.params.HttpConnectionParams;
 import org.apache.synapse.transport.http.conn.Scheme;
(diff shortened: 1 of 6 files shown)


abeykoon:
@arunans23 @nilminiwso2  these information should be captured at docs - will be important for administration.

arunans23:
Created a doc issue, to track the documentation progress. https://github.com/wso2/micro-integrator/issues/1604

PR #1484, opened 2020-02-10
Fix wso2/product-ei/issues/4961

malakaganga:
## Purpose
When the socket timeout exceeds in the REQUEST_DONE state TargetHandler invokes the targetErrorHandler and it invokes the SynapseCallbackRecievers receive method. But when this
happens the call back is already removed from the map. So the SynapseCallbackReciver prints the
 "Synapse received a response for the request with ... But a callback is not registered (anymore) to process this response" log.
This can occur at any time where targetErrorHandler invokes the SynapseCallbackRecievers receive method.
Fix this issue by differentiating the path by adding a state when SynapseCallbackReciever is invoked by targetErrorHandler

## Goals
Fixes: https://github.com/wso2/product-ei/issues/4961

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DeliveryAgent.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DeliveryAgent.java
@@ -146,7 +146,8 @@ public class DeliveryAgent {
                 }
                 if (queue.size() == maxWaitingMessages) {
                     MessageContext msgCtx = queue.poll();
-
+                    msgCtx.setProperty(PassThroughConstants.INTERNAL_EXCEPTION_ORIGIN,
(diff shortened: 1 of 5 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DeliveryAgent.java
```diff
@@ -185,6 +186,8 @@ public void errorConnecting(HttpRoute route, int errorCode, String message) {
 			MessageContext msgCtx = queue.poll();
 
 			if (msgCtx != null) {
+                msgCtx.setProperty(PassThroughConstants.INTERNAL_EXCEPTION_ORIGIN,
```
GDLMadushanka: shall we fix the formatting issue

PR #1505, opened 2020-03-30
Fixes the issue of tasks not getting removed when member count reduced

Nirothipan:
Fixes https://github.com/wso2/micro-integrator/issues/1303

When the synapse artifacts are updated ( for mp ) we are not removing the previously deployed tasks, hence the changes in the member count is not captured  and if the previous count is higher, the extra tasks still remains in the system.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
@@ -18,11 +18,6 @@
  */
 package org.apache.synapse.message.processor.impl;
 
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Properties;
-import java.util.concurrent.Callable;
-
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/message/processor/impl/ScheduledMessageProcessor.java
```diff
@@ -310,11 +310,20 @@ public void destroy(){
...
+
+	@Override
+	public void destroy(boolean preserveState , boolean isArtifactUpdate) {
 
 		if (!preserveState) {
-			stop();
 			deleteMessageProcessorState();
 		}
+		if (isArtifactUpdate || !preserveState) {
+			/*	All the tasks need to be stopped during artifact update and re deployed so that the changes in member
```
sdkottegoda: There seems to be an indentation issue

Nirothipan: This is because the previous one is intended with tabs, I tried formatting but it generates a big diff. However this shows up correctly in idea.

PR #1507, opened 2020-03-31
Add messages processor states to task store in Micro Integrator

Nirothipan:
With the fix for https://github.com/wso2/product-ei/issues/872, a new property is introduced and was written in registry to maintain the message processor states. 

However in Micro Integrator, we use a RDBMS based task store and to remove the registry sharing , we can write the same to that store and retrieve it. Hence introducing this fix . 

Fixes : https://github.com/wso2/micro-integrator/issues/1071

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 
 # ignore target directories
 target/
+modules/transports/core/vfs/repository/conf/*.properties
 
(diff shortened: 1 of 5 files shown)


Nirothipan:
Closed as sharing registry is needed and we will fix registry issue .

PR #1508, opened 2020-04-02
Fix MalformedJsonException when Source String is not a valid json

arunans23:
## Purpose
Resolves https://github.com/wso2/product-ei/issues/5034

Here we try to parse the source property and check whether it is a valid JSON. If not, we will continue on the existing XPath flow.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
@@ -282,15 +282,21 @@ public class Target {
                                      MessageContext synCtx) {
         if (action.equals(ACTION_REPLACE) && !sourceNodeList.isEmpty() && sourceNodeList.get(0) instanceof OMText) {
             String sourceString = ((OMText)sourceNodeList.get(0)).getText();
-            JsonElement jsonElement = jsonParser.parse(sourceString);
-            if (jsonElement instanceof JsonObject || jsonElement instanceof JsonArray) {
-                try {
-                    JsonUtil.getNewJsonPayload(((Axis2MessageContext) synCtx).getAxis2MessageContext(),
-                            sourceString, true, true);
-                    return;
-                } catch (AxisFault af) {
-                    log.error("Could not add json object to the json stream", af);
+            try {
+                JsonElement jsonElement = jsonParser.parse(sourceString);
+                if (jsonElement instanceof JsonObject || jsonElement instanceof JsonArray) {
+                    try {
+                        JsonUtil.getNewJsonPayload(((Axis2MessageContext) synCtx).getAxis2MessageContext(),
+                                sourceString, true, true);
+                        return;
+                    } catch (AxisFault af) {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
```diff
@@ -282,15 +282,21 @@ private void insertElementToBody(ArrayList<OMNode> sourceNodeList, OMElement e,
...
+                    try {
+                        JsonUtil.getNewJsonPayload(((Axis2MessageContext) synCtx).getAxis2MessageContext(),
+                                sourceString, true, true);
+                        return;
+                    } catch (AxisFault af) {
+                        log.error("Could not add json object to the json stream", af);
+                    }
                 }
+            } catch (JsonSyntaxException ex) {
+                synLog.traceOrDebug("Source property string is not a valid json");
```
GDLMadushanka, line 297: shall we add ex

PetteriPertola:
Hi @arunans23, is this fixed in WSO Integrator v 6.6.0?
I am still seeing this exactly in this use case.

arunans23:
@PetteriPertola This fix was included recently. It is not available in EI 6.6.0.

PetteriPertola:
@arunans23 thanks - two follow ups:
1. When will it be released in a production version of EI? 
2. Any known workaround? I am trying to format from XML to a plain text string that will be sent out as a fixed-length text message to a queue. Since the text does not have any json curly brackets around it, it fails at this point.

arunans23:
@PetteriPertola 
1. We currently do not have a release date for the next EI 6xx series. We will be doing a release in EI 7xx series during the beginning of the 3rd quarter. 
2. As a workaround, instead of using the Enrich mediator, you can use the PayloadFactory mediator to enrich the payload into the body.
```
<api xmlns="http://ws.apache.org/ns/synapse" name="check" context="/check">
   <resource methods="GET">
      <inSequence>
         <payloadFactory media-type="text">
            <format>            Check        </format>
            <args/>
         </payloadFactory>
         <property name="fixedLengthMessage" value="ABC DEFG" scope="default" type="STRING"/>
         <payloadFactory media-type="text">
            <format>$1</format>
            <args>
               <arg evaluator="xml" expression="$ctx:fixedLengthMessage"/>
            </args>
         </payloadFactory>
         <respond/>
      </inSequence>
   </resource>
</api>
          
```

PetteriPertola:
@arunans23 the payloadFactory trims the whitespace from both sides of the 'Check' message
Is there any way to stop it from doing that? 
I need the whitespace as it must be a fixed length outgoing message.

PetteriPertola:
In case anyone else faces the same issue - my workaround was to create a class mediator where I do the following:
String message = context.getEnvelope().getBody().getFirstElement().getText();
message = String.format("%-92s", message);
context.getEnvelope().getBody().getFirstElement().setText(message);

This adds whitespace to the end of the message making it exactly 92 characters long, thus re-adding the trimmed whitespace by PayloadFactory.

PR #1510, opened 2020-04-06
Add support for dynamic names in property mediator

sajithaliyanage:
Property mediator does not allow dynamic names at the moment. It is supported for other mediators - e.g. sequence mediator. This is a useful feature to have.

Examples:
```
<property name="{get-property('propertyName')}" expression="$ctx:propertyValue" />
<property name="{$ctx:propertyName}" expression="$ctx:propertyValue" />
<property name="{json-eval({$ctx:propertyName})}" expression="$ctx:propertyValue" />
```

From this PR, the property mediator can not have a static name like `{name}` since this uses the same logic used in defining the sequence mediator key attribute.

Related issue- https://github.com/wso2/product-ei/issues/3300


sajithaliyanage:
Check **isContentAware** variable for Property mediator when it has a dynamic name value - https://github.com/wso2/wso2-synapse/pull/1512

PR #1532, opened 2020-05-12
Preserve INTERNAL_TRANSACTION_COUNTED property in Passthru transport

sajinieKavindya:
## Purpose
To track the transactions in the Passthru transport (as part of the transaction counting mechanism), the message property called "INTERNAL_TRANSACTION_COUNTED" is maintained throughout the message flow.

Related PRs: https://github.com/wso2/wso2-axis2/pull/207, https://github.com/wso2/wso2-axis2-transports/pull/268

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -330,7 +330,11 @@
             <groupId>org.apache.axis2</groupId>
             <artifactId>axis2-transport-local</artifactId>
         </dependency>
-        <!--  JSON Schema validator dependency -->
+	<dependency>
+		<groupId>org.apache.axis2</groupId>
+		<artifactId>axis2-transport-base</artifactId>
+	</dependency>
+	<!--  JSON Schema validator dependency -->
         <dependency>
             <groupId>org.wso2.orbit.com.github.fge</groupId>
             <artifactId>json-schema-validator-all</artifactId>
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
@@ -37,6 +37,7 @@ import org.apache.axis2.description.AxisServiceGroup;
 import org.apache.axis2.description.TransportOutDescription;
 import org.apache.axis2.description.WSDL2Constants;
 import org.apache.axis2.engine.AxisConfiguration;
+import org.apache.axis2.transport.base.BaseConstants;
 import org.apache.axis2.transport.http.HTTPConstants;
 import org.apache.axis2.transport.http.HTTPTransportUtils;
 import org.apache.axis2.wsdl.WSDLConstants;
@@ -56,12 +57,12 @@ import org.apache.synapse.transport.passthru.config.PassThroughConfiguration;
 import org.apache.synapse.util.MediatorPropertyUtils;
 import org.apache.synapse.util.MessageHelper;
 
-import javax.mail.internet.ContentType;
-import javax.mail.internet.ParseException;
-import javax.xml.namespace.QName;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import javax.mail.internet.ContentType;
+import javax.mail.internet.ParseException;
+import javax.xml.namespace.QName;
 
 /**
  * This is a simple client that handles both in only and in out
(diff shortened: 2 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
```diff
@@ -159,6 +160,15 @@ public static void send(
         String preserveAddressingProperty = (String) synapseOutMessageContext.getProperty(
                 SynapseConstants.PRESERVE_WS_ADDRESSING);
         MessageContext axisOutMsgCtx = cloneForSend(originalInMsgCtx, preserveAddressingProperty);
+
+        // set "INTERNAL_TRANSACTION_COUNTED" property in axisOutMsgCtx's message context, if it is present in the message
+        // context of the originalInMsgCtx
+        Object transactionProperty = originalInMsgCtx.getProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED);
+        if (transactionProperty instanceof Boolean) {
```
sachithKay, line 167: Since this is an internally set property, do we really need to check instanceof?

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
```diff
@@ -159,6 +160,15 @@ public static void send(
...
+            axisOutMsgCtx.setProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED,
+                                      Boolean.getBoolean(transactionProperty.toString()));
```
sachithKay, line 169: are we setting the INTERNAL_TRANSACTION_COUNTED value as a System property?

modules/core/pom.xml
```diff
@@ -330,7 +330,11 @@
             <groupId>org.apache.axis2</groupId>
             <artifactId>axis2-transport-local</artifactId>
         </dependency>
-        <!--  JSON Schema validator dependency -->
+		<dependency>
```
sdkottegoda, line 333: Is there an indentation issue?

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
sdkottegoda: I think we don't need it. But we need to do a null check if we're removing this

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
```diff
@@ -384,7 +385,11 @@ private void handleMessage(String messageID ,MessageContext response,
                     axisOutMsgCtx.getProperty(org.apache.axis2.Constants.OUT_TRANSPORT_INFO));
             response.setTransportIn(axisOutMsgCtx.getTransportIn());
             response.setTransportOut(axisOutMsgCtx.getTransportOut());
-
+            // set "INTERNAL_TRANSACTION_COUNTED" property in response.
+            Object transactionProperty = axisOutMsgCtx.getProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED);
+            if (transactionProperty instanceof Boolean) {
```
sdkottegoda, line 390: We can remove the instance of check here aswell

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
```diff
@@ -384,7 +385,11 @@ private void handleMessage(String messageID ,MessageContext response,
...
+                response.setProperty(BaseConstants.INTERNAL_TRANSACTION_COUNTED, transactionProperty);
```
sdkottegoda, line 391: I think we can drop the response paths since we are only considering the request in flow?

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
sachithKay: null check and Boolean.valueOf should work.

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2FlexibleMEPClient.java
sajinieKavindya: No, we are not.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
```diff
@@ -513,7 +511,11 @@ public MessageContext createMessageContext(MessageContext msgContext, SourceRequ
                 conn.getContext().getAttribute(PassThroughConstants.CORRELATION_ID));
         msgContext.setProperty(PassThroughConstants.CORRELATION_LOG_STATE_PROPERTY,
                 sourceConfiguration.isCorrelationLoggingEnabled());
-
+        // propagate transaction property
+        Object transactionProperty = conn.getContext().getAttribute(BaseConstants.INTERNAL_TRANSACTION_COUNTED);
+        if (transactionProperty instanceof Boolean) {
```
sdkottegoda, line 516: same here

modules/core/pom.xml
sajinieKavindya: No indentation issues shown in the local IDE.

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
sajinieKavindya: Removing since we are not tracking the response path.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/ServerWorker.java
sajinieKavindya: Done.

modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
sajinieKavindya: Done.

PR #1542, opened 2020-05-24
Refactor rabbitmq store-forward implementation

indikasampath2000:
- update the amqp-client library to the latest version
- the producer changed to use the publisher confirms for reliable publishing
- the consumer revamp according to the latest amqp-client version

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
@@ -1,13 +1,13 @@
 /**
- *  Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
- *
- *  WSO2 Inc. licenses this file to you under the Apache License,
- *  Version 2.0 (the "License"); you may not use this file except
- *  in compliance with the License.
- *  You may obtain a copy of the License at
- *
- *    http://www.apache.org/licenses/LICENSE-2.0
- *
+ * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * <p>
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -1,13 +1,13 @@
...
- *
+ * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * <p>
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
abeykoon, line 10: Shouldn't we update the Copyright to 2020?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+    private Connection connection;
+    private Channel channel;
+    private RabbitMQStore store;
+    private String queueName;
+    private String idString;
+    private CachedMessage cachedMessage; // Holds the last message read from the message store
+
+    public RabbitMQConsumer(RabbitMQStore store) {
+        if (store == null) {
+            log.error("Cannot initialize.");
```
abeykoon, line 53: Cannot initialise what? Shall we say 
`Cannot initialise RabbitMQ message processor + message processor name`? 
My suggestion is to handle a situation using exceptions rather returning null. During deployment, this should fail the deployment (ultimately the CAR file deployment fails)

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+     * Get a single message from the queue and deserialize for set into the message context
+     *
+     * @return the {@link MessageContext} with received message
+     */
+    @Override
+    public MessageContext receive() {
+        if (isAlive()) {
+            try {
+                GetResponse delivery = channel.basicGet(queueName, false);
+                if (delivery != null) {
```
abeykoon, line 70: Can `delivery` be null? Shouldn't we handle that situation?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
-		return null;
-	}
+    /**
+     * Get a single message from the queue and deserialize for set into the message context
+     *
+     * @return the {@link MessageContext} with received message
+     */
+    @Override
+    public MessageContext receive() {
+        if (isAlive()) {
```
abeykoon, line 67: This does not check if message processor is live, rather it checks if connection is live. Shall we change the method name?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+    /**
+     * Cleanup the {@link Connection} and it's related resources
+     *
+     * @return whether cleanup successful or not
+     */
     @Override
-    public boolean isAlive() {
+    public boolean cleanup() {
+        if (connection != null) {
+            connection.abort();
```
abeykoon, line 135: what does this method do? Is it closing the connection and release the network resources? Otherwise there can be a connection leak here.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+                    if (log.isDebugEnabled()) {
+                        log.debug(getId() + " Received MessageId: " + delivery.getProps().getMessageId());
+                    }
+                    return synapseMc;
+                }
+            } catch (ShutdownSignalException | IOException e) {
+                log.error(getId() + " connection error when receiving messages." + e);
+            }
+        } else {
+            cleanup();
```
abeykoon, line 86: Let's say RabbitMQ is down temporarily. If MP is configured to receive 1 message per sec, this will execute reconnection logic in that much frequency. Is that desirable? Should we block execution, wait until connection becomes live again and start? Connection retry can be done on a separate thread.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+            setConnection(store.createConnection());
+            setChannel(store.createChannel(connection));
```
abeykoon, line 88: Creating a connection may throw exceptions. Are we handling them?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+     * @return a {@link StorableMessage} object
+     * @throws IOException
+     */
+    private StorableMessage deserializeMessage(Channel channel, GetResponse delivery) throws IOException {
+        StorableMessage storableMessage = null;
+        try (ByteArrayInputStream inputStream = new ByteArrayInputStream(delivery.getBody());
+             ObjectInput objectInput = new ObjectInputStream(inputStream)) {
+            storableMessage = (StorableMessage) objectInput.readObject();
+        } catch (ClassNotFoundException e) {
+            log.error(getId() + "unable to read the stored message." + e);
```
abeykoon, line 107: The log should carry which message store.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+            channel.basicAck(delivery.getEnvelope().getDeliveryTag(), false);
```
abeykoon, line 108: This is not something that needs to be handled by `deserializeMessage` method. Looking at the exception thrown, message consumer need to do this AFAIU.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+                    synapseMc = MessageConverter.toMessageContext(storableMessage, axis2Mc, synapseMc);
+                    updateCache(delivery, delivery.getProps().getMessageId());
+                    log.info(getId() + " Received MessageId: " + delivery.getProps().getMessageId());
+                    if (log.isDebugEnabled()) {
+                        log.debug(getId() + " Received MessageId: " + delivery.getProps().getMessageId());
+                    }
+                    return synapseMc;
+                }
+            } catch (ShutdownSignalException | IOException e) {
+                log.error(getId() + " connection error when receiving messages." + e);
```
abeykoon, line 83: Should not we cleanup the connection and channel upon this? This can be tricky. However, during half-open connection situation or due to any other undesirable network error, we may get an exception while receiving a message. Upon that I think we better reset.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+                }
+            }
+            return false;
+        }
+
+        public GetResponse getMessage() {
+            return message;
+        }
+
+        public CachedMessage setId(String id) {
```
abeykoon, line 243: Shall we move this to the constructor of CachedMessage?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+            this.message = message;
+        }
+
+        public boolean ack() {
+            if (message != null && channel != null && channel.isOpen()) {
+                try {
+                    channel.basicAck(message.getEnvelope().getDeliveryTag(), false);
+                    return true;
+                } catch (IOException e) {
+                    log.error(getId() + " cannot ack last read message. Error: " + e.getLocalizedMessage(), e);
```
abeykoon, line 232: We cannot ack using the channel. Maybe it is broken. Should we clean it up and reset? As I saw, if we could not ack,  we just retry the same message right?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+    private boolean isInitialized = false;
+    private String idString; // ID of the MessageProducer
+
+    /**
+     * The RabbitMQ producer
+     *
+     * @param store the {@link RabbitMQStore} object
+     */
+    public RabbitMQProducer(RabbitMQStore store) {
+        if (store == null) {
```
abeykoon, line 60: Same as consumer. If parameters are null shall we throw exceptions so that deployment fails with required message?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+            log.error("Cannot initialize.");
```
abeykoon, line 61: The log should carry message store name.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+
+    /**
+     * Store the given message into the queue and return whether the operation success or not
+     *
+     * @param synCtx Message to be saved.
+     * @return {@code true} if storing of the message is successful, {@code false} otherwise.
+     */
+    @Override
+    public boolean storeMessage(MessageContext synCtx) {
+        if (synCtx == null) {
```
abeykoon, line 76: Is this a check we need to do? What are the situations that this can become null?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+            return false;
+        }
+        if (connection == null) {
```
abeykoon, line 79: Message store and processor is used for guaranteed delivery scenarios. If we just keep shut up upon connection is null, the flow that is trying to store the message doesn't get that guarantee. Isn't it?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+            log.error(getId() + " cannot proceed. RabbitMQ Connection is null. Ignored MessageId: " +
+                    synCtx.getMessageID());
+            return false;
+        }
+        boolean result = false;
+        try (Channel channel = connection.createChannel()) {
```
abeykoon, line 85: I belive `connection.createChannel()` is not a heavy operation? Otherwise, this will effect the performance. We need to keep the channel in a variable and reuse.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+            channel.confirmSelect();
+            StorableMessage storableMessage = MessageConverter.toStorableMessage(synCtx);
+            final byte[] message = serializeMessage(storableMessage);
+            final AMQP.BasicProperties basicProperties = getBasicProperties(synCtx, storableMessage);
+            publishMessage(channel, exchangeName, routingKey, basicProperties, message);
+            result = channel.waitForConfirms();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
abeykoon, line 93: What happens if this occurs? We better add a log here. @Asitha

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
abeykoon: I see an exception unhandled on my IDE

image

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+     * @param routingKey      the routing key
+     * @param basicProperties other properties for the message
+     * @param messageBody     the message body
+     * @throws IOException
+     */
+    private void publishMessage(Channel channel, String exchangeName, String routingKey,
+                                AMQP.BasicProperties basicProperties, byte[] messageBody) throws IOException {
+        if (StringUtils.isNotEmpty(exchangeName)) {
+            channel.basicPublish(exchangeName, routingKey, basicProperties, messageBody);
+        } else {
```
abeykoon, line 150: We better add a debug  log here.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,186 @@
...
+    }
+
+    /**
+     * Used to close the channel opened in this object instance.
+     * This should be called after the end of each call on storeMessage method
+     * But instead of this, try with resources will close the channel
+     */
+    @Override
+    public boolean cleanup() {
+        return true;
```
abeykoon, line 162: Shouldn't we close the created rabbitmq connection here?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: The `RabbitMQConsumer.java` was already there with the license header. Therefore, just leave it as it is.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack to improve the error log message. The `return;` statement used not to initialize the constructor upon the store becomes null. That is how it was handled in other types of store processor as well. Technically, the store won't become null if you deploy the processor by creating through management console or tooling.

https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsConsumer.java#L66

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Yes. The `basicGet` doesn't block. It returns immediately and returns null if there are no messages in the queue.  https://www.rabbitmq.com/api-guide.html#getting

The `null` is returning at the end of the `receive` method. It handled by the top level.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: The `isAlive` method suppose to check the availability of connectivity with the message store. The processor must be deactivated if the `isAlive` returns false. That method inherited from the `MessageConsumer`.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: It supposes to close the connection and its channel forcefully.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: The default value of the retry count is 3 and the retry delay is 30000ms. The message processor supposes to connect within the specified values in a case of connection failure. The processor will deactivate if it couldn't connect.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Yes. RabbitMQStore#createConnection will handle exceptions.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: The `getId` has the store name.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack. Move it to the `receive` method.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: It is creating at the RabbitMQConsumer constructor. The messageId is unknown at that time.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: As per the implementation, the processor should be deactivated in case of a connection error with the broker. The processor implementation handled it based on the boolean value returned by the consumer. The `ack` method in the RabbitMQConsumer invoked by the processor upon successful backend invocation. The method must return whether it is successful or not.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: That is how it was handled in other types of store processor as well. Technically, the store won't become null if you deploy the processor by creating through management console or tooling.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: Not sure. It was there in the older implementation and other message producers' implementation as well.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: The store mediator invokes the fault sequence in such a case. The client has to retry based on the response sent in the fault sequence. The mediation engine doesn't suppose to retry internally.

modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
```diff
@@ -70,25 +76,32 @@ public static void init() throws NoSuchFieldException {
...
+        temp.put(RabbitMQStore.RETRY_COUNT, "0");
         rabbitMQStore.setParameters(temp);
         AxisConfiguration axisConfiguration = new AxisConfiguration();
         configurationContext = new ConfigurationContext(axisConfiguration);
         SynapseConfiguration synapseConfiguration = new SynapseConfiguration();
         Axis2SynapseEnvironment env = new Axis2SynapseEnvironment(configurationContext, synapseConfiguration);
-        rabbitMQStore.init(env);
+        try {
+            rabbitMQStore.init(env);
+        } catch (Exception ignore) {
```
sdkottegoda, line 88: What are we trying to assert here? If it's the 'UnknownHostException', let's assert that. catching 'Exception will hide any exception that's thrown, hence make the test case not usefull

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+
+    private Connection connection;
+    private Channel channel;
+    private RabbitMQStore store;
+    private String queueName;
+    private String idString;
+    private CachedMessage cachedMessage; // Holds the last message read from the message store
+
+    public RabbitMQConsumer(RabbitMQStore store) {
+        if (store == null) {
```
sdkottegoda, line 52: Can the store ever be null? It's called inside RabbitMQStore.getConsumer() passing (this)

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
sdkottegoda: The store can never be null right? It's called inside RabbitMQStore.getConsumer() passing (this). In that case, we can remove the if block?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: No. This method should be called after the end of each call on storeMessage method. But it doesn't require since try with resources used.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: Do you mean to the else condition or after `if-else` ladder? The `else` block is obvious since the exchange name not given.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
sdkottegoda, line 86: Do we need to log an error here?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
```diff
@@ -34,255 +34,220 @@
...
+        if (isAlive()) {
+            try {
+                GetResponse delivery = channel.basicGet(queueName, false);
+                if (delivery != null) {
+                    StorableMessage storableMessage = deserializeMessage(channel, delivery);
+                    org.apache.axis2.context.MessageContext axis2Mc = store.newAxis2Mc();
+                    MessageContext synapseMc = store.newSynapseMc(axis2Mc);
+                    synapseMc = MessageConverter.toMessageContext(storableMessage, axis2Mc, synapseMc);
+                    updateCache(delivery, delivery.getProps().getMessageId());
+                    log.info(getId() + " Received MessageId: " + delivery.getProps().getMessageId());
```
sdkottegoda, line 76: Do we need this log, which we'll get per message?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
sdkottegoda, line 107:
```suggestion
            log.error(getId() + " unable to read the stored message. ", e);
```

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
sdkottegoda, line 83:
```suggestion
                log.error(getId() + " connection error when receiving messages.", e);
```

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
sdkottegoda, line 108: Should we acknowledge upon a ClassNotFound? which means that we cannot process the message?

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: As per the store mediator design, a channel has to create and close per message.

`produceStatus = messageStore.getProducer().storeMessage(newCtx);`

The above error might be due to rabbitmq client library not updated properly.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
```diff
@@ -36,147 +36,202 @@
...
+            Thread.currentThread().interrupt();
+        } catch (TimeoutException | IOException e) {
+            String errorMsg = getId() + ". Ignored MessageId: " + synCtx.getMessageID() + ". " +
+                    "Could not store message to store [" + store.getName() + "]. " +
+                    "Error:" + e.getLocalizedMessage();
+            log.error(errorMsg, e);
+        }
+        if (log.isDebugEnabled()) {
+            log.debug(getId() + ". Stored MessageId: " + synCtx.getMessageID());
+        }
```
sdkottegoda, line 109: I think this has to be inside the try block. The log message will be confusing after an error situation.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Yes. It was there in other implementation as well. There could be a situation where somebody is copying message processor synapse config to the deployment without messages store synapse config. I guess it won't make any harm just keeping that as it is.

modules/core/src/test/java/org/apache/synapse/message/store/RabbitMQStoreTest.java
indikasampath2000: I tried to do something similar to `@Test(expected = SynapseException.class)`. But `@BeforeClass` annotation doesn't support that. Therefore, catch the exception to make sure init doesn't fail.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack. I added it for testing purpose and forgot to remove.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQProducer.java
indikasampath2000: Ack.

modules/core/src/main/java/org/apache/synapse/message/store/impl/rabbitmq/RabbitMQConsumer.java
indikasampath2000: Technically, it should be a message where you directly publish to the queue without store mediator. In such a case, logging an exception and acknowledge the message.

PR #1543, opened 2020-05-26
Modify REST API identification to avoid duplicate identification

Nipunich:
## Purpose
Modify REST API identification to avoid duplicate API identification when the Prometheus Observability Handler is executed.


modules/core/src/main/java/org/apache/synapse/rest/RESTConstants.java
```diff
@@ -85,4 +85,10 @@
                                                                     _SYNAPSE_INTERNAL_+ "Access-Control-Allow-Headers";
     public static final String INTERNAL_CORS_HEADER_ORIGIN = _SYNAPSE_INTERNAL_+ "Origin";
 
+    /**
+     * Constant related to identifying API when Prometheus Observability Handler is present
+     */
+    public static final String IS_ALREADY_PROCESSED = "IS_ALREADY_PROCESSED";
```
sdkottegoda, line 91: How about IS_PROMETHEUS_ENGAGED? Just a suggestion because 'IS_ALREADY_PROCESSED' seems a bit too generic.

modules/core/src/main/java/org/apache/synapse/rest/API.java
```diff
@@ -59,6 +60,7 @@
     private Map<String,Resource> resources = new LinkedHashMap<String,Resource>();
     private List<Handler> handlers = new ArrayList<Handler>();
     private String swaggerResourcePath;
+    private RESTUtils restUtils;
```
sdkottegoda, line 63: Util classes are usually classes that hold static util methods. Is this initialization essential?

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -153,4 +153,17 @@ private static void handleException(String msg, Throwable t) {
         throw new SynapseException(msg, t);
     }
 
+    /**
+     * Identify the API using the API context.
+     *
+     * @param path    request path
+     * @param context API context
+     */
+    public boolean getRESTApi(String path, String context) {
```
sdkottegoda, line 162: this can be a static util method, can't it?

sdkottegoda: Also I would think the method 'getRestApi' returns an api path, but apparently it returns a boolean indicating whether the the path is a rest api? Or what does it return? Shall we return the method accordingly?

modules/core/src/main/java/org/apache/synapse/rest/RESTConstants.java
arunans23: Shouldn't this constant match with what is in Micro-Integrator https://github.com/wso2/micro-integrator/blob/8d6762305b7f61716fcb6c929e7387467d454587/components/mediation/data-publishers/org.wso2.micro.integrator.prometheus.observability/src/main/java/org/wso2/micro/integrator/prometheus/observability/metrics/util/PrometheusMetricsConstants.java#L44

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
```diff
@@ -149,4 +141,36 @@ private void apiProcessNonDefaultStrategy(MessageContext synCtx, API api) {
             api.process(synCtx);
         }
     }
+
+    public boolean identifyAPI(API api, MessageContext synCtx) {
+        Collection<API> apiSet = synCtx.getEnvironment().getSynapseConfiguration().getAPIs();
+        List<API> defaultStrategyApiSet = new ArrayList<API>(apiSet);
```
arunans23, line 147: Shall we get this list directly from the parent method?

arunans23: Also shall we make it private?

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
```diff
@@ -75,32 +75,24 @@ private boolean dispatchToAPI(MessageContext synCtx) {
...
-                    || api.getVersionStrategy().getClass().getName().equals(URLBasedVersionStrategy.class.getName())) {
+        if (null != synCtx.getProperty(RESTConstants.IS_ALREADY_PROCESSED)) {
+            API api = (API) synCtx.getProperty("PROCESSED_API");
+            identifyAPI(api, synCtx);
+        } else {
+            for (API api : apiSet) {
+                identifyAPI(api, synCtx);
+            }
+
+            for (API api : defaultStrategyApiSet) {
```
arunans23, line 87: can't we combine this and the above for loop into one loop?

modules/core/src/main/java/org/apache/synapse/rest/API.java
```diff
@@ -90,6 +92,7 @@
     public API(String name, String context) {
         super(name);
         setContext(context);
+        restUtils = new RESTUtils();
```
arunans23, line 95: Can't we use a static function? here

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/API.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/API.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/RESTConstants.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
Nipunich: As they loop over two different sets. So we can't combine them.

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -153,4 +153,17 @@ private static void handleException(String msg, Throwable t) {
         throw new SynapseException(msg, t);
     }
 
+    /**
+     * Identify the API using the API context.
+     *
+     * @param path    request path
+     * @param context API context
+     */
```
arunans23, line 161: Add the doc for return param here

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
```diff
@@ -153,4 +153,17 @@ private static void handleException(String msg, Throwable t) {
...
+    public static boolean isRESTApiPath(String path, String context) {
```
arunans23, line 162: IMHO, the method name is not explanatory enough. Javadoc is also not clear. Shall we change that?

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
arunans23: Having two for loops here is costly. We should do something here to eliminate one.

Nipunich: It was the same two loops we had from the initial implementation of the RESTRequestHandler Class.

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/rest/RESTUtils.java
Nipunich: Fixed

tishan89:
@arunans23 are there any more changes in this PR that needs to be addressed?

PR #1544, opened 2020-05-27
Fix latency issue which happens due to HostName resolving

Vathsan:
## Purpose
This PR fixes the latency issue which happens due to HostName resolving.
We only need to resolve a hostname if we have configured a custom SSL profile. Because we retrieve the SSL profile configuration from `sslByHostMap` using the hostname. Otherwise, we do not need to resolve the hostname. 

Fixes https://github.com/wso2/product-ei/issues/4723


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/ClientConnFactory.java
```diff
@@ -81,8 +81,17 @@ public ClientConnFactory(
     
     private SSLContext getSSLContext(final IOSession iosession) {
         InetSocketAddress address = (InetSocketAddress) iosession.getRemoteAddress();
-        String host = address.getHostName() + ":" + address.getPort();
-        return  getSSLContextForHost(host);
```
Nirothipan, line 85: Shall we pass the address variable to getSSLContextForHost method and modify it there . It seems that is being used in some other place too.

PR #1545, opened 2020-05-28
Masking JMS Message Store parameter values using secure-vault

Nipunich:
## Purpose
Implement the Masking JMS Message Store parameter values using secure-vault capability.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
+++ b/modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
@@ -124,7 +124,7 @@ public class JmsStore extends AbstractMessageStore {
     /** regex for secure vault expression */
     private static final String SECURE_VAULT_REGEX = "\\{(wso2:vault-lookup\\('(.*?)'\\))\\}";
 
-    private Pattern vaultLookupPattern = Pattern.compile(SECURE_VAULT_REGEX);
+    private Pattern queueLookupPattern = Pattern.compile(SECURE_VAULT_REGEX);
 
     public MessageProducer getProducer() {
         if (cacheLevel == 1 && cachedProducer != null) {
@@ -494,7 +494,7 @@ public class JmsStore extends AbstractMessageStore {
             Object value = e.getValue();
             if (value instanceof String) {
                 if (CONNECTION_STRING.equals(e.getKey())) {
-                    value = resolveQueueName(value);
+                    value = resolveConnectionStringValues(value);
                 }
                 connectionProperties.put(e.getKey(), value);
             }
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -712,4 +723,17 @@ private String nameString() {
     public void setProducer(MessageProducer producer) {
         this.producer = producer;
     }
+
+    public String resolveQueueName(Object param) {
```
sdkottegoda, line 727: I feel like this method is to resolve all the vault-look ups in the connection string. In that case, why did we name this as 'resolveQueueName'?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -119,6 +121,11 @@
 
     private SynapseEnvironment synapseEnvironment;
 
+    /** regex for secure vault expression */
+    private static final String SECURE_VAULT_REGEX = "\\{(wso2:vault-lookup\\('(.*?)'\\))\\}";
+
+    private Pattern vaultLookupPattern = Pattern.compile(SECURE_VAULT_REGEX);
```
sdkottegoda, line 127: Can we use the the variables defined in SecureVaultResolver without re-defining them?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
sdkottegoda, line 727: I feel like this method is to resolve all the vault-look ups in the connection string. In that case, why did we name this as 'resolveQueueName'?

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
```diff
@@ -49,6 +49,8 @@
 import java.util.NoSuchElementException;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class JmsStore extends AbstractMessageStore {
```
arunans23, line 55: Are we doing this only for JmsStore? What about other message stores?

Nipunich:
The testing is in progress. Will reopen once completed.

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
Nipunich: Fixed

modules/core/src/main/java/org/apache/synapse/message/store/impl/jms/JmsStore.java
Nipunich: At the moment we did this only for JMS Store.

arunans23: Let's rename the PR title as JMS Message Store.

Nipunich: Done

PR #1549, opened 2020-06-08
Add Look ahead class validation while deserialization

sajinieKavindya:
## Purpose
This PR will avoid the insecure deserialization of Storable objects. 

## Security Concern
When type checking is performed on a deserialized object as in this code, there is a chance that an attacker can instantiate any object available to the application with an arbitrary internal state. In other words, an attacker could instantiate any class on the classpath with malicious member variables they wanted. This would allow an attacker to cause remote code execution upon deserialization or create malicious deserialized objects that can cause remote code execution and data tampering upon usage. Hence, this is a threat.

### DIFF
--- a/modules/securevault/src/main/java/org/apache/synapse/securevault/tool/CipherTool.java
+++ b/modules/securevault/src/main/java/org/apache/synapse/securevault/tool/CipherTool.java
@@ -21,6 +21,7 @@ package org.apache.synapse.securevault.tool;
 import org.apache.commons.cli.*;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.commons.util.LookAheadObjectInputStream;
 import org.apache.synapse.securevault.*;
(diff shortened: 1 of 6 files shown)


sachithKay:
Is this only happening in JDBCMessageStores? 
I think same could happen in other places like RabbitMQConsumer as well.

modules/securevault/src/main/java/org/apache/synapse/securevault/tool/CipherTool.java
```diff
@@ -387,6 +388,7 @@ private static Key getKey(String filePath) {
         ObjectInputStream in = null;
         try {
             in = new ObjectInputStream(new FileInputStream(keyFile));
+            in = new LookAheadObjectInputStream(new FileInputStream(keyFile), Key.class);
```
Nirothipan, line 391: do we still nee line 390

sajinieKavindya: Removed.

PR #1566, opened 2020-06-19
Fix issue in destroying mediator configurations upon re-deployment of proxy service

Vathsan:
[1] https://github.com/wso2/wso2-synapse/blob/29597f5967e2d123e79dd487b2613f6bf59a887f/modules/core/src/main/java/org/apache/synapse/deployers/APIDeployer.java#L123

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
+++ b/modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
@@ -159,6 +159,7 @@ public class ProxyServiceDeployer extends AbstractSynapseArtifactDeployer {
                 }
                 ProxyService currentProxy = getSynapseConfiguration().getProxyService(existingArtifactName);
                 currentProxy.stop(getSynapseConfiguration());
+                currentProxy.destroy();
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/deployers/ProxyServiceDeployer.java
```diff
@@ -159,6 +159,7 @@ public String updateSynapseArtifact(OMElement artifactConfig, String fileName,
                 }
                 ProxyService currentProxy = getSynapseConfiguration().getProxyService(existingArtifactName);
                 currentProxy.stop(getSynapseConfiguration());
+                currentProxy.destroy();
```
Nirothipan, line 162: should we destroy at line 118 also ?

Nirothipan: and this seems not destroyed properly in APIDeployer undeploySynapseArtifact method too , shall we fix that too with this.

Vathsan: +1, fixed both in d17210e81d6224de123b303528a9e1cca5da0c1b

PR #1581, opened 2020-06-29
Remove Json stream after aggregation

madushadhanushka:
Removing the JSON stream after aggregated using XML path to fix inconsistent behavior in logging the payload.
Fix wso2/product-ei#5160


modules/core/src/main/java/org/apache/synapse/mediators/eip/aggregator/AggregateMediator.java
```diff
@@ -619,6 +620,11 @@ private MessageContext getAggregatedMessage(Aggregate aggregate) {
             } catch (AxisFault axisFault) {
                 log.error("Error occurred while setting the new JSON payload to the msg context", axisFault);
             }
+        } else {
+            // Removing the JSON stream after aggregated using XML path.
+            // This will fix inconsistent behaviour in logging the payload.
+            ((Axis2MessageContext) newCtx).getAxis2MessageContext()
+                                          .removeProperty(Constants.ORG_APACHE_SYNAPSE_COMMONS_JSON_JSON_INPUT_STREAM);
```
arunans23, line 627: Shall we use the JsonUtil method here? https://github.com/wso2/wso2-synapse/blob/master/modules/commons/src/main/java/org/apache/synapse/commons/json/JsonUtil.java#L957

PR #1588, opened 2020-07-14
Fix Synapse handler chain called twice during an API invocation

Vathsan:
## Purpose
In https://github.com/wso2/wso2-synapse/pull/1543, we have refactored and moved the existing logic inside identifyAPI() method. In the earlier logic, we return a boolean response in dispatchToAPI() method. But after refactoring we do not pass the return value of identifyAPI() method to dispatchToAPI() method. Hence the mediation flow continues after executing identifyAPI() method. Which is not the expected behaviour. This PR fixes the issue.

Fixes https://github.com/wso2/micro-integrator/issues/1839

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
@@ -78,19 +78,12 @@ public class RESTRequestHandler {
         API defaultAPI = null;
         if (null != synCtx.getProperty(RESTConstants.IS_PROMETHEUS_ENGAGED)) {
             API api = (API) synCtx.getProperty(RESTConstants.PROCESSED_API);
-            identifyAPI(api, synCtx, defaultStrategyApiSet);
+            if (identifyAPI(api, synCtx, defaultStrategyApiSet)) {
+                return true;
+            }
         } else {
             for (API api : apiSet) {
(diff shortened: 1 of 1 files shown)


sdkottegoda:
Shall we describe what was the root cause for the issue and how this fix resolves it?

Vathsan:
> Shall we describe what was the root cause for the issue and how this fix resolves it?

Updated the desccription with more details.

modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
```diff
@@ -159,15 +156,25 @@ private boolean identifyAPI(API api, MessageContext synCtx, List defaultStrategy
...
+            return new States(defaultAPI, true);
         }
-        return false;
+        return new States(defaultAPI, false);
+    }
+}
+
+class States {
+    API defaultAPI;
+    boolean state;
```
sdkottegoda, line 174: state = true/false is misleading. e.g., I don't understand the meaning of `if (states.state)' at line 89. Can we have a better wording?

Vathsan: Shall we go with 'isApiProcessed' instead of state?

Vathsan: Resolving this as we have modified the fix.

PR #1589, opened 2020-07-15
Fix MI-1526: Error when enriching String property with two words to Json payload

chanikag:
## Purpose
This is to fix https://github.com/wso2/micro-integrator/issues/1526

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/elementary/EnrichMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/elementary/EnrichMediator.java
@@ -142,8 +142,19 @@ public class EnrichMediator extends AbstractMediator {
                     isSourcePropertyXML = true;
                 }
             } catch (JsonSyntaxException e) {
-                synLog.traceOrDebug("Source string is not a valid json");
-                isSourcePropertyXML = true;
+                try {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/elementary/EnrichMediator.java
```diff
@@ -142,8 +142,19 @@ public boolean mediate(MessageContext synCtx) {
                     isSourcePropertyXML = true;
                 }
             } catch (JsonSyntaxException e) {
-                synLog.traceOrDebug("Source string is not a valid json");
-                isSourcePropertyXML = true;
+                try {
+                    // Enclosing using quotes due to the following issue
+                    // https://github.com/google/gson/issues/1286
+                    String enclosedSourceProperty = "\"" + sourceProperty + "\"";
```
sdkottegoda, line 148: Is there any harm in doing this inside the try itself, rather than enclosing with " inside the catch after it fails once?

chanikag: Then there can be a possibility to enclose json objects, json arrays..etc.

PR #1612, opened 2020-07-30
cache API object in order to use in future usages

tharindu1st:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


modules/core/src/main/java/org/apache/synapse/rest/RESTRequestHandler.java
```diff
@@ -158,6 +170,7 @@ private boolean identifyAPI(API api, MessageContext synCtx, List defaultStrategy
             if (log.isDebugEnabled()) {
                 log.debug("Located specific API: " + api.getName() + " for processing message");
             }
+            synCtx.setProperty(RESTConstants.PROCESSED_API, api);
```
isudana, line 173: shall we move this to process() method of the API so that we don't need to set it over and over again here into different cases?

tharindu1st: done

PR #1617, opened 2020-08-05
Add property values to the Jaeger tracing

malakaganga:
## Purpose
In Jaeger Tracing span tags Before mediation and after mediation are the same. If we put a payload factory mediator and change the payload, in span tags of PayloadFactoryMediator it shows the same payload for beforePayload tag
and the afterPayload tag. Fix this issue by changing the design of close event handling.

Also Currently, in EI Jager tracing, we can only see the property name. This commit improves the code to add the property value as well.

## Goals
Fixes: https://github.com/wso2/product-ei/issues/5218
Fixes: https://github.com/wso2/product-ei/issues/5215

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/opentracing/management/helpers/SpanTagger.java
+++ b/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/opentracing/management/helpers/SpanTagger.java
@@ -19,9 +19,6 @@
 package org.apache.synapse.aspects.flow.statistics.opentracing.management.helpers;
 
 import io.opentracing.Span;
-import org.apache.synapse.MessageContext;
-import org.apache.synapse.aspects.flow.statistics.data.raw.BasicStatisticDataUnit;
-import org.apache.synapse.aspects.flow.statistics.data.raw.StatisticDataUnit;
 import org.apache.synapse.aspects.flow.statistics.data.raw.StatisticsLog;
 import org.apache.synapse.aspects.flow.statistics.opentracing.OpenTracingManagerHolder;
 import org.apache.synapse.aspects.flow.statistics.opentracing.models.SpanWrapper;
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
```diff
@@ -145,6 +150,15 @@ public boolean mediate(MessageContext synCtx) {
...
                 synCtx.setProperty(name, resultValue);
 
+            } else if (XMLConfigConstants.SCOPE_TRACE.equals(scope)) {
+                //Setting property value into the propertyValue variable for tracing purposes
+                if (resultValue != null ) {
+                    if (resultValue instanceof OMElement) {
+                        ((OMElement) resultValue).build();
+                    }
+                    //Converted to string since only used to display as a span tag
+                    propertyValue = resultValue.toString();
```
Vathsan, line 160: Following this approach can we add multiple property values to a single span?

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/opentracing/management/helpers/SpanTagger.java
```diff
@@ -40,37 +37,45 @@ private SpanTagger() {}
...
+                    if (spanWrapper.getCloseEventStatisticDataUnit().getContextPropertyMap() != null) {
+                        span.setTag("afterContextPropertyMap",
+                                spanWrapper.getCloseEventStatisticDataUnit().getContextPropertyMap().toString());
                     }
+                } else if (openStatisticsLog.getContextPropertyMap() != null) {
+                    span.setTag("afterContextPropertyMap", openStatisticsLog.getContextPropertyMap().toString());
+                }
+                if (spanWrapper.getCloseEventStatisticDataUnit() != null &&
+                        spanWrapper.getCloseEventStatisticDataUnit().getPropertyValue() != null) {
+                    span.setTag("propertyMediatorValue",
```
Vathsan, line 70: Should not we need to add the property name with the value while setting the span tag?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/PropertyMediator.java
malakaganga: This is to support adding the value of each property mediator which is in trace scope to span tags previously only property mediator name was added.

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/opentracing/management/helpers/SpanTagger.java
malakaganga: The property name is already being added

PR #1623, opened 2020-08-24
Avoid calculate message length for get request

madushadhanushka:
Avoid calculating message length for GET request as it should be always zero.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
@@ -374,7 +374,9 @@ public class PassThroughHttpSender extends AbstractHandler implements TransportS
                 long messageSize = 0;
 
                 try {
-                    OverflowBlob overflowBlob = setStreamAsTempData(formatter, msgContext, format);
+                    boolean hasNoMessageBody = HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(
+                            Constants.Configuration.HTTP_METHOD)) || RelayUtils.isDeleteRequestWithoutPayload(msgContext);
+                    OverflowBlob overflowBlob = setStreamAsTempData(formatter, msgContext, format, hasNoMessageBody);
                     messageSize = overflowBlob.getLength();
                     msgContext.setProperty(PassThroughConstants.PASSTROUGH_MESSAGE_LENGTH, messageSize);
                     deliveryAgent.submit(msgContext, epr);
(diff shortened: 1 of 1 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
```diff
@@ -383,16 +383,18 @@ private void sendRequestContent(final MessageContext msgContext, final EndpointR
...
                         if (HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(Constants.Configuration.HTTP_METHOD)) ||
-                                RelayUtils.isDeleteRequestWithoutPayload(msgContext)) {
+                            RelayUtils.isDeleteRequestWithoutPayload(msgContext)) {
                             pipe.setSerializationCompleteWithoutData(true);
-                        } else if (messageSize == 0 &&
+                            return;
+                        }
+                        overflowBlob.writeTo(out);
+                        if (pipe.isStale) {
+                            throw new IOException("Target Connection is stale..");
```
Nirothipan, line 395: Don't we need to check whether pipe is stale or not before using it in line 390

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
```diff
@@ -383,16 +383,18 @@ private void sendRequestContent(final MessageContext msgContext, final EndpointR
...
                     if (out != null) {
-                        overflowBlob.writeTo(out);
-                        if (pipe.isStale) {
-                            throw new IOException("Target Connection is stale..");
-                        }
                         //if HTTP MEHOD = GET we need to write down the HEADER information to the wire and need
                         //to ignore any entity enclosed methods available.
                         if (HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(Constants.Configuration.HTTP_METHOD)) ||
-                                RelayUtils.isDeleteRequestWithoutPayload(msgContext)) {
+                            RelayUtils.isDeleteRequestWithoutPayload(msgContext)) {
```
Nirothipan, line 389: This condition should already be evaluated in line 377 when calling setStreamAsTempData .. can we assign it to a variable before calling it and re use here. Else this will be done two times in some flows, wdyt

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpSender.java
```diff
@@ -374,7 +374,9 @@ private void sendRequestContent(final MessageContext msgContext, final EndpointR
                 long messageSize = 0;
 
                 try {
-                    OverflowBlob overflowBlob = setStreamAsTempData(formatter, msgContext, format);
+                    boolean hasBodyContent = HTTPConstants.HTTP_METHOD_GET.equals(msgContext.getProperty(
```
indikasampath2000, line 377: Shall we rename the variable name `hasBodyContent` as `hasMessageBody`?

indikasampath2000 (dismissed):

indikasampath2000 (approved):

PR #1628, opened 2020-09-09
Facilitate inner get-property in a XPathFunction.

sachithKay:
This fix modifies the fix introduced in 22278f6.
Fixes: wso2/product-ei#5242

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
@@ -68,8 +68,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.StringTokenizer;
-
-
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseXPath.java
```diff
@@ -214,22 +214,25 @@ public SynapseXPath(String xpathString) throws JaxenException {
...
+        if (xPathString.contains("get-property")) {
             // extract property args
-            xpath = xPathString.substring(xPathString.indexOf("(") + 1, xPathString.indexOf(")"));
-	    args = xpath.split(",");
-	    if (args != null && args.length == 2) {
-		// remove white space
-		scope = args[0].trim();
-		// remove opening and ending quotes
-		scope = scope.substring(1, scope.length() - 1);
+            Matcher matcher = Pattern.compile("get-property\\(([^)]+)\\)").matcher(xPathString);
```
GDLMadushanka, line 221: can we use string maniulation insted of regex here
since we are already inside if (xPathString.contains("get-property")) { check and regex is expensive

GDLMadushanka (approved):

PR #1633, opened 2020-09-25
Support renaming the key name in json payloads

dinuish94:
## Approach

- The new key value must be specified as a string in the source.

Supported types for source are **inline, property and custom**.

- Target should contain the xpath to locate the key in the below format.

Consider the following payload.

```
{
	“user”: {
		“name”: “Kevin”,
		“age”: 26
         }
}
```

If you want to rename the ‘name’ attribute to ‘firstName’ the xpath expression should be json-eval(‘$.user.name’)

- Target type should be **key** and target action should be **replace**.

E.g.: xml
```
<enrich>
      <source clone="true" type="inline"> firstName </source>
      <target action="replace" type="key" xpath="json-eval($.user.name)"/>
</enrich>

```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
@@ -498,6 +498,25 @@ public class Target {
     }
 
     /**
+     * Renames a json key name at the specified json path with a new key name
+     *
+     * @param synapseContext Current message context
+     * @param jsonPath       The path to locate the key. Should be resolved to a map or an array including map items.
+     * @param keyName        Current name of the key
+     * @param newKeyName     New name of the key
+     * @throws IOException if failed to set the new json payload
+     */
+    public void renameKey(MessageContext synapseContext, String jsonPath, String keyName, String newKeyName)
+            throws IOException {
+
+        org.apache.axis2.context.MessageContext axis2MessageCtx =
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
```diff
@@ -497,6 +497,25 @@ public void removeJson(MessageContext synCtx, SynapsePath jsonPath) throws IOExc
...
+     * @param newKeyName     New name of the key
+     * @throws IOException
+     */
+    public void renameKey(MessageContext synapseContext, String jsonPath, String keyName, String newKeyName)
+            throws IOException {
+
+        org.apache.axis2.context.MessageContext axis2MessageCtx =
+                ((Axis2MessageContext) synapseContext).getAxis2MessageContext();
+        String payload = JsonUtil.jsonPayloadToString(((Axis2MessageContext) synapseContext).getAxis2MessageContext());
+        DocumentContext doc = JsonPath.parse(payload).renameKey(jsonPath, keyName, newKeyName);
```
tishan89, line 515: Do we have to get the string and then again parse it? Can't we treat it as json

dinuish94: `renameKey` method has to be executed on an object of type `DocumentContext`, hence we will have to parse it.

GDLMadushanka (dismissed):

GDLMadushanka (approved):

PR #1634, opened 2020-10-02
Introduce new JSON type to property

GDLMadushanka:
properties
Fixes wso2/product-ei/issues/5266

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
@@ -45,10 +45,16 @@ import org.jaxen.JaxenException;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class SynapseJsonPath extends SynapsePath {
 
     private static final Log log = LogFactory.getLog(SynapseJsonPath.class);
+
+    private static final String EXTRACT_PROP_REGEX = "^(\\$ctx|\\$trp|\\$axis2):([a-zA-Z0-9]+)";
+
+    private String propertyExpression;
 
     private String enableStreamingJsonPath = SynapsePropertiesLoader.loadSynapseProperties().
     getProperty(SynapseConstants.STREAMING_JSONPATH_PROCESSING);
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
```diff
@@ -91,7 +98,22 @@ public SynapseJsonPath(String jsonPathExpression)  throws JaxenException {
         EIPUtils.setJsonPathConfiguration();
 
         this.contentAware = true;
-        this.expression = jsonPathExpression;
+        // supporting evaluation against a property
+        // Ex: json-eval($ctx:prop1.student.name)
+        Pattern fullPattern = Pattern.compile(JSON_PATH_WITH_PROP_REGEX);
```
tishan89, line 103: Shall we to a small test to compare this with Startswith()

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
```diff
@@ -125,6 +147,20 @@ public String stringValueOf(MessageContext synCtx) {
...
         }
+
+        // evaluating the jsonPath against a property
+        if (propertyExpression != null) {
+            try {
+                SynapseXPath xPath = new SynapseXPath(propertyExpression);
+                String result = xPath.stringValueOf(synCtx);
+                return stringValueOf(result);
+            } catch (JaxenException e) {
+                log.error("Xpath evaluation failed", e);
```
tishan89, line 158: log the expression too

tishan89 (changes requested):

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
GDLMadushanka: +1 fixed

modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseJsonPath.java
GDLMadushanka: +1 fixed

tishan89 (approved):

PR #1638, opened 2020-10-06
Refactor checkXMLVersion in PayloadFactory to improve performance

arunans23:
## Purpose

Resolves wso2/product-ei#5279

This PR refactors the PayloadFactory mediator code to improve the performance in the following ways.

- Initialize Document Builder only once.
- Turn off unnecessary features in DocumentBuilder
- Check XML version for every request. (Not for every argument in PF)
- Use stringToOM as a last resort inside isXML() check.


arunans23:
Closing this PR, since there are performance issues with this approach. (In high Core environments)

PR #1650, opened 2020-10-20
Improve enrich remove action to support properties

GDLMadushanka:
Fixes wso2/product-ei/issues/5199

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
@@ -296,6 +296,26 @@ public class EnrichMediatorFactory extends AbstractMediatorFactory {
         return -1;
     }
 
+    // Adding this method to be consistent with convertTypeToInt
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
```diff
@@ -296,6 +296,25 @@ private int convertTypeToInt(String type) {
         return -1;
     }
 
+    private int convertActionToInt(String action) {
```
tishan89, line 299: Why do we need to convert enums to ints

tishan89 (changes requested):

modules/core/src/main/java/org/apache/synapse/config/xml/EnrichMediatorFactory.java
GDLMadushanka: I followed the previous style used in the code ( there's a similar func for source/target types called convertTypeToInt )

tishan89 (approved):

PR #1653, opened 2020-10-26
Freemarker support for PayloadFactory

tharakamd:
## Purpose

This PR adds FreeMarker support for the PayloadFactory mediator. A new attribute is added to the synapse syntax of the PayloadFactory mediator to switch between default script implementation and FreeMarker implementation. For backward compatibility, if that attribute is not set, then the default implementation would get selected. 

## Samples
A sample synapse syntax is as bellow,

```xml
<payloadFactory media-type="xml" template-type="freemarker">
                <format><![CDATA[<name>${payload.name}</name>]]></format>
                <args/>
</payloadFactory>
```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/PayloadFactoryMediatorSerializer.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/PayloadFactoryMediatorSerializer.java
@@ -91,14 +91,14 @@ public class PayloadFactoryMediatorSerializer extends AbstractMediatorSerializer
                     OMElement formatElem = fac.createOMElement(FORMAT, synNS);
                 String type = mediator.getType();
                 if(type!=null && (type.contains(JSON_TYPE) || type.contains(TEXT))) {
-                     formatElem.setText(mediator.getFormat());
+                    if (isFreeMarkerTemplate(mediator)) {
+                        createCdataTag(mediator,formatElem);
+                    }else{
+                        formatElem.setText(mediator.getFormat());
+                    }
                 } else {
                     if (isFreeMarkerTemplate(mediator)) {
-                        String formatString = mediator.getFormat();
-                        formatString = removeCDATAFromPayload(formatString);
-                        OMTextImpl omText = (OMTextImpl) formatElem.getOMFactory().createOMText(formatElem, formatString,
-                                XMLStreamConstants.CDATA);
-                        formatElem.addChild(omText);
+                        createCdataTag(mediator, formatElem);
                     } else {    
                     formatElem.addChild(AXIOMUtil.stringToOM(mediator.getFormat()));
                 }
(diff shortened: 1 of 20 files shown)


modules/core/src/main/java/org/apache/synapse/config/xml/PayloadFactoryMediatorSerializer.java
```diff
@@ -148,8 +163,23 @@ public OMElement serializeSpecificMediator(Mediator m) {
...
+        return mediator.getTemplateType() != null && mediator.getTemplateType().equalsIgnoreCase(FREEMARKER);
+    }
     public String getMediatorClassName() {
         return PayloadFactoryMediator.class.getName();
     }
 
+    public static String removeCDATAFromPayload(String inputPayload) {
+        if (inputPayload.startsWith("<![CDATA[")) {
+            inputPayload = inputPayload.substring(9);
+            int i = inputPayload.indexOf("]]>");
```
GDLMadushanka, line 177: since the index of returns, the first occurrence is there a chance this will break in CDATA inside CDATA scenario?

modules/core/src/main/java/org/apache/synapse/mediators/transform/pfutils/FreeMarkerTemplateProcessor.java
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * Inject a JSON payload in the FreeMarker
+     *
+     * @param messageContext Message context
+     * @param data           FreeMarker data input
+     */
+    private void injectJsonPayload(Axis2MessageContext messageContext, Map<String, Object> data) {
+
+        org.apache.axis2.context.MessageContext axis2MessageContext = messageContext.getAxis2MessageContext();
+        String jsonPayloadString = JsonUtil.jsonPayloadToString(axis2MessageContext);
```
GDLMadushanka, line 218: Can this be a JSONPrimitive ex:- following are valid JSON payloads 
boolean - true or false
single number
single string
null

modules/core/src/main/java/org/apache/synapse/config/xml/PayloadFactoryMediatorSerializer.java
tharakamd: fixed this

modules/core/src/main/java/org/apache/synapse/mediators/transform/pfutils/FreeMarkerTemplateProcessor.java
tharakamd: fixed this using gson and checking for JSON primitives.

GDLMadushanka (approved):

abeykoon:
When adding documentation, please add a xml based example, json based example and an example with CDATA. Make sure to update payloadFactory syntax at master doc branch.

tharakamd:
> When adding documentation, please add a xml based example, json based example and an example with CDATA. Make sure to update payloadFactory syntax at master doc branch.

Sure, I will add those to the documentation.

GDLMadushanka (dismissed):

GDLMadushanka (approved):

PR #1654, opened 2020-10-27
Fix the errors thrown in enrich mediator xpath json-eval 

Nipunich:
## Purpose
Fix the errors thrown when the Enrich mediator is used with In-line source and xpath="json-eval.." target property.

Fixes wso2/product-ei#5257

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Source.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Source.java
@@ -30,11 +30,14 @@ import org.apache.axiom.soap.SOAPEnvelope;
 import org.apache.axiom.soap.SOAPFactory;
 import org.apache.axiom.soap.SOAPHeaderBlock;
 import org.apache.axiom.soap.impl.builder.StAXSOAPModelBuilder;
+import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


GDLMadushanka (approved):

modules/core/src/main/java/org/apache/synapse/mediators/elementary/Source.java
```diff
@@ -269,6 +274,20 @@ public JsonElement evaluateJson(MessageContext synCtx, SynapseLog synLog,
         JsonParser parser = new JsonParser();
         if (xpath != null) {
             SynapseJsonPath sourceJsonPath = (SynapseJsonPath) this.xpath;
+
+            if (InlineExpressionUtil.checkForInlineExpressions(sourceJsonPath.toString())) {
+                try {
+                    String jsonpath = sourceJsonPath.toString();
+                    if (jsonpath.startsWith("json-eval(")) {
```
abeykoon, line 281: are there json paths those do not start with "json-eval("?

abeykoon: OK. Seems json-path is ultimately an XPATH. :) Please correct me if wrong.

PR #1662, opened 2020-11-13
Add capability to define source and target for call mediator

malakaganga:
## Purpose
Fixes: https://github.com/wso2/product-ei/issues/5296

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
@@ -20,29 +20,36 @@
 package org.apache.synapse.mediators.builtin;
 
 import org.apache.axis2.AxisFault;
+import org.apache.axis2.Constants;
+import org.apache.axis2.context.ConfigurationContext;
+import org.apache.axis2.context.ConfigurationContextFactory;
 import org.apache.synapse.ContinuationState;
 import org.apache.synapse.ManagedLifecycle;
 import org.apache.synapse.MessageContext;
 import org.apache.synapse.SynapseConstants;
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/builtin/CallMediator.java
```diff
@@ -190,9 +266,38 @@ private boolean handleBlockingCall(MessageContext synInCtx) {
             log.error("Error while performing the call operation in blocking mode");
             return false;
         }
+        postMediate(synInCtx, originalMessageType, originalContentType, originalTransportHeaders);
         return true;
     }
 
+    public void postMediate(MessageContext response, String originalMessageType, String originalContentType,
+                            Map originalTransportHeaders) {
+        if (isTargetAvailable()) {
```
GDLMadushanka, line 275: shall we use isTargetAvailable for consistency

modules/core/src/main/java/org/apache/synapse/util/CallMediatorEnrichUtil.java
```diff
@@ -0,0 +1,299 @@
...
+import org.apache.synapse.mediators.elementary.EnrichMediator;
+import org.apache.synapse.mediators.elementary.Source;
+import org.apache.synapse.mediators.elementary.Target;
+import org.apache.synapse.transport.passthru.util.RelayUtils;
+import org.jaxen.JaxenException;
+
+import javax.xml.namespace.QName;
+import java.util.ArrayList;
+import java.util.Map;
+
```
GDLMadushanka, line 53: shall we add class level comment

GDLMadushanka (approved):

PR #1671, opened 2020-12-06
Do not update Context state if the current state is the same

giafar:
If `-DenableCorrelationLogs=true` is set and the endpoint called responds in a chunked way both `org.apache.synapse.transport.passthru.TargetContext`and `org.apache.synapse.transport.passthru.SourceContext` are called for each chunk and an entry in the log is written: this behavior can be responsable of slowdowns because of logger is not async or file system performance.
 
## Purpose
In a pass through request, if -DenableCorrelationLogs=true is set and the endpoint response is chunked avoid to write correlation log for each RESPONSE_BODY updateState to increase performance.

## Goals
Fix `org.apache.synapse.transport.passthru.TargetContext`and `org.apache.synapse.transport.passthru.SourceContext` just to return immediatly if the state parameter has the same value of the connection context.


isudana (approved):

isudana:
Thank you for your contribution.

PR #1683, opened 2021-01-05
Add Stream Interceptor 

Nirothipan:
1. whenever the request data stream enters the engine from the client
 2. whenever the request data stream is written to the client
 3. whenever the response data stream enters from back to the engine
 4. whenever the response data stream is written to the client

These will be engaged via stream-interceptors.xml file found in conf directory.

```xml
  <interceptors xmlns:svns="http://org.wso2.securevault/configuration">
      <interceptor class="org.apache.synapse.transport.passthru.SampleStreamInterceptor">
          <parameter name="charset" value="ISO-8859-1"/>
          <parameter name="enableInterception" value="true"/>
      </interceptor>
  </interceptors>
```

In addition to the above for methods, the interceptors will have additional 4 methods to enable engagement of the corresponding methods.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
@@ -21,8 +21,8 @@ package org.apache.synapse.transport.passthru;
 import org.apache.axis2.context.MessageContext;
 
 import java.nio.ByteBuffer;
-import java.util.HashMap;
 import java.util.Map;
+import java.util.Properties;
 
 /**
  * Default class implementing the methods of {@link StreamInterceptor}. A stream interceptor can be written by
@@ -31,8 +31,7 @@ import java.util.Map;
  */
 public abstract class DefaultStreamInterceptor implements StreamInterceptor {
 
-    private String name;
-    private Map<String, Object> properties = new HashMap<String, Object>();
+    private Properties properties = new Properties();
 
     public boolean interceptSourceRequest(MessageContext axisCtx) {
         return false;
@@ -74,11 +73,4 @@ public abstract class DefaultStreamInterceptor implements StreamInterceptor {
         return properties;
     }
 
-    public String getName() {
-        return name;
-    }
-
-    public void setName(String name) {
-        this.name = name;
-    }
 }
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.synapse.transport.passthru;
+
+import org.apache.axis2.context.MessageContext;
+
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Default class implementing the methods of {@link StreamInterceptor}. A stream interceptor can be written by
+ * extending and overding the necessary methods of this class.
+ * For an example see{@link SampleStreamInterceptor}
+ */
+public abstract class DefaultStreamInterceptor implements StreamInterceptor {
+
+    private String name;
+    private Map<String, Object> properties = new HashMap<String, Object>();
+
+    public boolean interceptSourceRequest(MessageContext axisCtx) {
+        return false;
+    }
+
+    public boolean sourceRequest(ByteBuffer buffer, MessageContext axisCtx) {
+        return true;
+    }
+
+    public boolean interceptTargetRequest(MessageContext axisCtx) {
+        return false;
+    }
+
+    public void targetRequest(ByteBuffer buffer, MessageContext axisCtx) {
+
+    }
+
+    public boolean interceptTargetResponse(MessageContext axisCtx) {
+        return false;
+    }
+
+    public boolean targetResponse(ByteBuffer buffer, MessageContext axisCtx) {
+        return true;
+    }
+
+    public boolean interceptSourceResponse(MessageContext axisCtx) {
+        return false;
+    }
+
+    public void sourceResponse(ByteBuffer buffer, MessageContext axisCtx) {
+
+    }
+
+    public void addProperty(String name, Object value) {
+        properties.put(name, value);
+    }
+
+    public Map getProperties() {
+        return properties;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+}
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SampleStreamInterceptor.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SampleStreamInterceptor.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.synapse.transport.passthru;
+
+import org.apache.axis2.context.MessageContext;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+
+/**
+ * Sample stream interceptor, which intercepts source and target request to print the stream passing via the engine.
+ * <pre>
+ * {@code
+ * <interceptors xmlns:svns="http://org.wso2.securevault/configuration">
+ *     <interceptor name="SampleStreamInterceptor" class="org.apache.synapse.transport.passthru.SampleStreamInterceptor">
+ *         <parameter name="charset" value="ISO-8859-1"/>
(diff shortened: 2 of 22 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/AbstractStreamInterceptor.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Abstract class implementing the methods of {@link StreamInterceptor}. A stream interceptor can be written by
+ * extending and overding the necessary methods of this class.
+ * For an example see{@link SampleStreamInterceptor}
+ */
+public abstract class AbstractStreamInterceptor implements StreamInterceptor {
+
+    private String name;
+    private Map<String, Object> properties = new HashMap<String, Object>();
```
sdkottegoda, line 35: Would java.util.Properties be a better pick? Can concurrency be an issue here?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
```diff
@@ -0,0 +1,158 @@
...
+    private static boolean isLoadedAlready = false;
+
+    /**
+     * Load and get all synapse interceptors
+     *
+     * @return List of loaded synapse interceptors
+     */
+    public static List<StreamInterceptor> getInterceptors() {
+
+        if (!isLoadedAlready) {
```
sdkottegoda, line 60: PassthruHttpListener.init being called concurrently? If so, this method get called concurrently and the list of interceptors will be inconsistent.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
```diff
@@ -0,0 +1,158 @@
...
+                } else {
+                    handleException("Name not defined in one or more interceptor");
+                }
+
+                if (interceptorElem.getAttribute(CLASS_Q) != null) {
+                    String className = interceptorElem.getAttributeValue(CLASS_Q);
+                    if (!"".equals(className)) {
+                        StreamInterceptor interceptor = createInterceptor(className);
+                        interceptors.add(interceptor);
+                        interceptor.setName(name);
```
sdkottegoda, line 92: To me, it seems as if the name cannot change in the runtime once the interceptor is created. Why don't we remove the setter and add the parameter to the constructor itself?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
```diff
@@ -0,0 +1,158 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
arunans23, line 2: Shall we change the licence year?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetHandler.java
```diff
@@ -86,14 +89,27 @@
...
+
+    public TargetHandler(DeliveryAgent deliveryAgent, ClientConnFactory connFactory,
+                         TargetConfiguration configuration) {
+        this(deliveryAgent, connFactory, configuration, new ArrayList<StreamInterceptor>());
+    }
+
     public TargetHandler(DeliveryAgent deliveryAgent,
                          ClientConnFactory connFactory,
-                         TargetConfiguration configuration) {
+                         TargetConfiguration configuration,
```
chanikag, line 103: Can this constructor is used by outside components? if so shall we use a new constructor with new parameters rather than doing an API change

modules/transports/core/nhttp/src/main/resources/stream-interceptors.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied. See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+<!--resides in conf folder-->
+<interceptors>
+    <interceptor name="SampleStreamInterceptor" class="org.apache.synapse.transport.passthru.SampleStreamInterceptor">
```
chanikag, line 20: If a sample interceptor is added by default for every request, it will check whether the interceptor is enabled

modules/transports/core/nhttp/src/main/resources/stream-interceptors.xml
```diff
@@ -0,0 +1,24 @@
...
+        <parameter name="charset" value="ISO-8859-1"/>
+        <parameter name="enableInterception" value="true"/>
```
chanikag, line 22: Why we need a global config to enable? And do we enable the sample interceptor by default?

modules/transports/core/nhttp/src/main/resources/stream-interceptors.xml
Nirothipan: no, we aren't enabling this. just added this in resources for reference

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/AbstractStreamInterceptor.java
Nirothipan: fixed

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
sdkottegoda: :D Consequences of coding at the end of the year...

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
Nirothipan: added a lock around this

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
Nirothipan: used a lock around this

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
Nirothipan: updated : p

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetHandler.java
Nirothipan: another constructor is defined already

modules/transports/core/nhttp/src/main/resources/stream-interceptors.xml
Nirothipan: just a sample implementation, user can refer this. This is not added to the conf.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
```diff
@@ -0,0 +1,76 @@
...
+    public void sourceResponse(ByteBuffer buffer, MessageContext axisCtx) {
+
+    }
+
+    public void addProperty(String name, Object value) {
+        properties.put(name, value);
+    }
+
+    public Map getProperties() {
+        return properties;
```
sdkottegoda, line 73: I was thinking, should we return an unmodifiable list? Or should we expose a method only to return a property specified by a key?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
```diff
@@ -0,0 +1,159 @@
...
+                                "StreamInterceptor parameter must contain name and value attributes, or a name and a child XML fragment");
+                    }
+                }
+            }
+        }
+    }
+
+    private static void handleException(String msg) {
+        log.error(msg);
+        throw new RuntimeException(msg);
```
sdkottegoda, line 152: Is throwing a runtime option a good selection here? The exception will get ignored unless specifically caught, and will cause the server to go to a dead state.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
```diff
@@ -0,0 +1,159 @@
...
+            OMElement child = (OMElement) it.next();
+
+            String propName = child.getAttribute(NAME_ATT).getAttributeValue();
+            if (propName == null) {
+                handleException("StreamInterceptor parameter must contain the name attribute");
+            } else {
+                if (child.getAttribute(VALUE_ATT) != null) {
+                    String value = child.getAttribute(VALUE_ATT).getAttributeValue();
+                    interceptor.addProperty(propName, value);
+                    PropertyHelper.setInstanceProperty(propName, value, interceptor);
```
sdkottegoda, line 135: Will this work? There isn't a method in the Interceptor class to as 'setPropName' right?

Nirothipan: yes this would, this will find the setter and set this value, same way it works in class mediators.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/util/StreamInterceptorsLoader.java
Nirothipan: yes, server is expected to go to a dead state and to be restarted since this is a config issue

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SampleStreamInterceptor.java
```diff
@@ -0,0 +1,111 @@
...
+ * <interceptors xmlns:svns="http://org.wso2.securevault/configuration">
+ *     <interceptor class="org.apache.synapse.transport.passthru.SampleStreamInterceptor">
+ *         <parameter name="charset" value="ISO-8859-1"/>
+ *         <parameter name="enableInterception" value="true"/>
+ *     </interceptor>
+ * </interceptors>
+ * }
+ * </pre>
+ */
+public class SampleStreamInterceptor extends DefaultStreamInterceptor {
```
sdkottegoda, line 41: Why don't we give it a proper name, such as 'LoggingStreamInterceptor' or something? Because, I feel like we will start asking customers to add this to debug issues in support in the near future. :)

sdkottegoda (approved):

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/DefaultStreamInterceptor.java
Nirothipan: done in https://github.com/wso2/wso2-synapse/pull/1695

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SampleStreamInterceptor.java
Nirothipan: updated in https://github.com/wso2/wso2-synapse/pull/1695

PR #1687, opened 2021-01-07
Upgrade HTTPComponents

arunans23:
This PR upgrades the HTTP components to the latest version.
Httpcore/Httpcore-nio : 4.4.14
Httpclient : 4.5.13

Other modifications: 
- Annotations have been modified according to https://github.com/apache/httpcomponents-core/commit/9e065bad07c9ca771c42e5b4f1dc12118c5e75c9
- Two test cases have been slightly modified to mock HttpRequest
- Clearing write events after output is produced. (Earlier httpcore-nio did this. Now they are supporting pipelining. Hence we have to do it in synapse side)


Nirothipan (approved):

arunans23:
This PR was reverted because of a bundle resolving issue in APIM. Same has been added with https://github.com/wso2/wso2-synapse/pull/1712

PR #1688, opened 2021-01-08
Add synpase xpath functions and iterations support for clone mediator

chanikag:
## Purpose
This PR is to add
1. synpase xpath functions for
- HMAC signature generation,
- Keystore based encryption,
- Keystore based decryption

2. Dynamic iterations support for clone mediator

### HMAC Generator - XPATH function

This XPath function computes a Hash-based message authentication code (HMAC) using a secret key. A HMAC is a small set of data that helps authenticate the origin of the message; it protects the integrity and the authenticity of the message.

Syntax:

hmac-generate(payload, secretKey, algorithm)

Example:

hmac-generate($body, ‘thisisasamplekey’)

| Parameter      | Description.                                                                      | Required/Optional                   |
| ------------- | ----------------------------------------------------- | ----------------------------- |
| payload          | The payload that has to be signed                                 | Required                                   |
| secretKey      | The secret Key                                                                  | Required                                   |
| algorithm       | The algorithm used to generate the HMAC signature   | Optional [Default : HmacSHA1] |

### Encrypt - XPATH function

This Xpath function encrypts the plain text using keystore using Asymmetric keys

Syntax: 

encrypt(plaintext, keystore_location, keystore_password, key_alias, keystore_type, algorithm)

Example:

encrypt(‘thisisasampletext’,$ctx:keystore,'wso2carbon', 'wso2carbon')

| Parameter      | Description.   | Required/Optional|
| ------------- | ------------- | ------------- | 
| plaintext  | The plain text to be encrypted | Required |
| keystore_location  | The keystore file location   | Required |
| keystore_password  | The keystore password  | Required |
| key_alias  | The alias of the certificarte to generate (private/public) key  | Required |
| keystore_type  | The keystore type  | Optional [Default : JKS] |
| algorithm  | The algorithm used to encrypt  | Optional [Default : RSA] |

### Decrypt - XPATH function

This Xpath function decrypt the cipher text using keystore using Asymmetric keys

Syntax: 

decrypt(plaintext, keystore_location, keystore_password, key_alias, keystore_type, algorithm)

Example:

decrypt('gNh9dTL3uL7gyB24QxVhXO9rRUPHsmQF7EFhEJNdUlUIFpksdmMuQD0KAn2pH7sF4JveScKPUUi31N8twwuqBefD7YBFv3Qh/hnJj//sGzAUJGfIs3eDn6okkNPrd76iO0EaQioJLCNDTYOc0RQDtyX0b2fAkZdpfbmBK7FxYIW1VE792WmfI/AaNqV1z4S/l5zLKisYEl0bVchmDj0WMF/JNM68YAlrCvf3QrgA+WleZWJikeV3mTRB5UcxAT8BBm8asbMMBHR9cg4TvxyerxhTcmk2kPdS5VCWmD7WI/WwZLRFOuKNOFlI5U8tV0WjbN+zYQt+k3cyR6oSSmBHqQ==', $ctx:keystore, 'wso2carbon',  'wso2carbon')

| Parameter      | Description.   | Required/Optional|
| ------------- | ------------- | ------------- | 
| plaintext  | The plain text to be encrypted | Required |
| keystore_location  | The keystore file location   | Required |
| keystore_password  | The keystore password  | Required |
| key_alias  | The alias of the certificarte to generate (private/public) key  | Required |
| keystore_type  | The keystore type  | Optional [Default : JKS] |
| algorithm  | The algorithm used to encrypt  | Optional [Default : RSA] |

### Clone mediator - dynamic iterations support

The Clone Mediator can be used to clone a message into several messages. In the current behavior it makes multiple identical copies of the message and execute the pre-defined set of targets. With this change, when the iterations attribute is defined, the clone mediator will clone message and execute the target according to the number of iterations iteratively. 

Syntax:

 <clone iterations="10">
<target>
      ----------------------
</target>
</clone> 

Examples:

iterations="10"
iterations="{get-property('propertyName')}"
iterations="{$ctx:propertyName}"
iterations="{json-eval({$ctx:propertyName})}"

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
@@ -67,6 +67,7 @@ public class CloneMediatorFactory extends AbstractMediatorFactory {
     private static final QName ID_Q
             = new QName(XMLConfigConstants.NULL_NAMESPACE, "id");
     private static final QName SEQUENTIAL_Q = new QName("sequential");
+    private static final String ITERATIONS = "iterations";
 
     /**
      * This method implements the createMediator method of the MediatorFactory interface
@@ -104,7 +105,6 @@ public class CloneMediatorFactory extends AbstractMediatorFactory {
             }
             mediator.setIterations(iterationsAttributeStr);
         }
-
 
         OMAttribute synchronousExeAttr= elem.getAttribute(SEQUENTIAL_Q);
         if (synchronousExeAttr != null && synchronousExeAttr.getAttributeValue().equals("true")) {
@@ -151,7 +151,7 @@ public class CloneMediatorFactory extends AbstractMediatorFactory {
         // ValueFactory for creating dynamic Value
         ValueFactory nameValueFactory = new ValueFactory();
         // create dynamic Value based on OMElement
-        return nameValueFactory.createValue("iterations", elem);
+        return nameValueFactory.createValue(ITERATIONS, elem);
     }
 
     /**
--- a/modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
+++ b/modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
@@ -23,8 +23,12 @@ import org.apache.axiom.om.OMAttribute;
 import org.apache.axiom.om.OMElement;
 import org.apache.axis2.util.JavaUtils;
 import org.apache.synapse.Mediator;
+import org.apache.synapse.mediators.Value;
 import org.apache.synapse.mediators.eip.Target;
 import org.apache.synapse.mediators.eip.splitter.CloneMediator;
+import org.apache.synapse.util.xpath.SynapseJsonPath;
+import org.apache.synapse.util.xpath.SynapseXPath;
+import org.jaxen.JaxenException;
 
 import javax.xml.namespace.QName;
 import java.util.Iterator;
@@ -35,7 +39,7 @@ import java.util.Properties;
  * different message contexts and mediated using the specified targets
  *
  * <pre>
- * &lt;clone [continueParent=(true | false)]&gt;
+ * &lt;clone [continueParent=(true | false)] [iterations="number"]&gt;
  *   &lt;target [to="uri"] [soapAction="qname"] [sequence="sequence_ref"]
  *          [endpoint="endpoint_ref"]&gt;
  *     &lt;sequence&gt;
@@ -56,6 +60,7 @@ public class CloneMediatorFactory extends AbstractMediatorFactory {
     private static final QName CLONE_Q
             = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "clone");
     private static final QName ATT_CONTINUE_PARENT = new QName("continueParent");
+    private static final QName ATT_ITERATIONS = new QName("iterations");
     private static final QName TARGET_Q
             = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "target");
 
@@ -90,6 +95,17 @@ public class CloneMediatorFactory extends AbstractMediatorFactory {
                     continueParent.getAttributeValue()));
         }
 
+        OMAttribute iterations = elem.getAttribute(ATT_ITERATIONS);
+        if (iterations != null) {
+            String iterationsAttributeStr = iterations.getAttributeValue();
+            if (isDynamicValue(iterationsAttributeStr)) {
+                Value iterationsAttributeValue = resolveDynamicAttribute(iterationsAttributeStr, elem);
+                mediator.setDynamicIterationsValue(iterationsAttributeValue);
+            }
+            mediator.setIterations(iterationsAttributeStr);
+        }
+
+
         OMAttribute synchronousExeAttr= elem.getAttribute(SEQUENTIAL_Q);
         if (synchronousExeAttr != null && synchronousExeAttr.getAttributeValue().equals("true")) {
         	asynchronousExe = false;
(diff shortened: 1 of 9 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/CloneMediator.java
```diff
@@ -135,6 +149,32 @@ public boolean mediate(MessageContext synCtx) {
...
+            ContinuationStackManager.addReliantContinuationState(clonedMsgCtx, i, getMediatorPosition());
+            target.mediate(clonedMsgCtx);
+        }
+    }
+
+    private int resolveIterationsCount(MessageContext synCtx) {
+        String countStr = "";
+        if (getDynamicIterationsValue() != null) {
+            countStr = getDynamicIterationsValue().evaluateValue(synCtx);
+        } else if (getIterations() != null) {
```
Nirothipan, line 172: do we need an else if, isn't only else is enough. i.e if getDynamicIterationsValue() == null, we should get from the other.

modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
```diff
@@ -90,6 +95,17 @@ public Mediator createSpecificMediator(OMElement elem, Properties properties) {
...
+        if (iterations != null) {
+            String iterationsAttributeStr = iterations.getAttributeValue();
+            if (isDynamicValue(iterationsAttributeStr)) {
+                Value iterationsAttributeValue = resolveDynamicAttribute(iterationsAttributeStr, elem);
+                mediator.setDynamicIterationsValue(iterationsAttributeValue);
+            }
+            mediator.setIterations(iterationsAttributeStr);
+        }
+
+
```
Nirothipan, line 108: shall we remove this empty line

modules/core/src/main/java/org/apache/synapse/util/xpath/DecryptFunction.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    public Object call(Context context, List args) throws FunctionCallException {
+        boolean debugOn = log.isDebugEnabled();
+
+        if (args == null) {
+            if (debugOn) {
+                log.debug("Missing arguments in the function call");
+            }
+            return SynapseXPathConstants.NULL_STRING;
+        }
```
Nirothipan, line 55: we can skip this if condition as this will anyway return null in the end

modules/core/src/main/java/org/apache/synapse/util/xpath/EncryptFunction.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public Object call(Context context, List args) throws FunctionCallException {
+        boolean debugOn = log.isDebugEnabled();
+        if (args == null) {
+            if (debugOn) {
+                log.debug("Missing arguments in the function call");
+            }
+            return SynapseXPathConstants.NULL_STRING;
+        }
```
Nirothipan, line 53: same here

modules/core/src/main/java/org/apache/synapse/util/xpath/EncryptFunction.java
```diff
@@ -0,0 +1,123 @@
...
+            if (log.isDebugEnabled()) {
+                log.debug("Certificate used for encrypting : " + certificate);
+            }
+            cipher.init(Cipher.ENCRYPT_MODE, certificate.getPublicKey());
+            byte[] cipherText = cipher.doFinal(plainTextBytes);
+            if (log.isDebugEnabled()) {
+                log.debug(String.format("Successfully encrypted data using the algorithm '%s'", algorithm));
+            }
+            return Base64.getEncoder().encodeToString(cipherText);
+
```
Nirothipan, line 117: shall we remove this empty line

modules/core/src/main/java/org/apache/synapse/util/xpath/HMACGenerateFunction.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Override
+    public Object call(Context context, List args) throws FunctionCallException {
+        boolean debugOn = log.isDebugEnabled();
+        if (args == null) {
+            if (debugOn) {
+                log.debug("Missing arguments in the function call");
+            }
+            return SynapseXPathConstants.NULL_STRING;
+        }
```
Nirothipan, line 49: shall we skip this if

modules/core/src/main/java/org/apache/synapse/util/xpath/KeyStoreManager.java
```diff
@@ -0,0 +1,102 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
+
+/*
+This is a util class for managing keystore
+ */
+
```
Nirothipan, line 37: shall we remove this empty line

modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
```diff
@@ -98,17 +114,68 @@ public Mediator createSpecificMediator(OMElement elem, Properties properties) {
...
+            } else {
+                new SynapseXPath(valueExpression);
+            }
+        } catch (JaxenException e) {
+            handleException("Invalid expression for attribute 'name' : " + attributeValue);
+        }
+        // ValueFactory for creating dynamic Value
+        ValueFactory nameValueFactory = new ValueFactory();
+        // create dynamic Value based on OMElement
+        return nameValueFactory.createValue("iterations", elem);
```
Nirothipan, line 154: shall we use a const for "iterations", since used in many places

modules/core/src/main/java/org/apache/synapse/util/xpath/DecryptFunction.java
chanikag: This is to avoid NPE when we do args.size()

modules/core/src/main/java/org/apache/synapse/util/xpath/EncryptFunction.java
chanikag: This is to avoid NPE when we do args.size()

modules/core/src/main/java/org/apache/synapse/util/xpath/HMACGenerateFunction.java
chanikag: This is to avoid NPE when we do args.size()

modules/core/src/main/java/org/apache/synapse/config/xml/CloneMediatorFactory.java
chanikag: This is used in this place only. However I have added a string constant in this class.

modules/core/src/main/java/org/apache/synapse/mediators/eip/splitter/CloneMediator.java
chanikag: +1

Nirothipan (approved):

PR #1689, opened 2021-01-08
Fix issue in http_gw.log rotation in windows

npamudika:
## Purpose
This PR fixes https://github.com/wso2/product-apim/issues/9560

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
@@ -275,6 +275,8 @@ public class AccessLogger {
                         createFileInError(dir, existing);
                     }
                 }
+		pathname = dir.getAbsolutePath() + File.separator + AccessConstants.getPrefix()
+                        + "_" + dateStamp + AccessConstants.getSuffix();
             }
             writer = new PrintWriter(new BufferedWriter(new FileWriter(
                     pathname, true), 128000), true);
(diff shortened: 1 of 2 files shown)


isharac (dismissed):

npamudika:
@Nirothipan Can you please review and merge this?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
```diff
@@ -275,6 +275,8 @@ protected synchronized void open() {
                         createFileInError(dir, existing);
                     }
                 }
+		pathname = dir.getAbsolutePath() + File.separator + AccessConstants.getPrefix()
+                        + "_" + dateStamp + AccessConstants.getSuffix();
```
Nirothipan, line 279: AFAIR, we couldn't come to a proper solution for this issue due to the windows environment and this fails when we rename the file. So shall we add this code block only if renameTo fails. Also, the inclusion of "_" is not bit clear, shall we define a constant for it or see how it is handled in String oldDateStamp and use the same way.

npamudika: Removed the underscore as per the offline discussion and tested the behaviour in linux too. We need to update the docs guiding to use the **access_log_prefix** ad **http_gw_** as we don't add an underscore in the code.

Nirothipan (approved):

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessLogger.java
npamudika: The changes to the http_gw log file rotation will work similar to http_access log file rotation and adds the date format to the log files.

PR #1691, opened 2021-01-13
Adding implementation for Inbound APIs

senthuran16:
## Purpose
Introducing Internal APIs to Synapse, and moving classes from `org.apache.synapse.rest` to `org.apache.synapse.api` to support this. (Commit for moving the classes: https://github.com/wso2/wso2-synapse/pull/1691/commits/446daf5879c312a86b3fc974e01e8a24fb6d123d).

### Introducing Internal APIs to Synapse
An Internal API is an API, which can only be invoked through a specified set of inbound endpoints. To define this, an API will have a `binds-to` attribute specified at either API level or Resource level. This attribute denotes that, from which inbound endpoint callers, can the API be invoked. 
- If `binds-to` is not specified at a Resource level, that Resource will inherit `binds-to` from the API level.
- If `binds-to` is specified at Resource level, that will override what has been provided in API level `binds-to`.
- `binds-to` of a Resource must be a subset of `binds-to` of its API.
- If `binds-to` is not present in an API, that API will be bound to `default`. This is then used when dispatching requests from various sources. (`default` can be stated as a binding too)

Refer to the examples below:

**Example 1: `binds-to` specified only at API level**
```xml
<api xmlns="http://ws.apache.org/ns/synapse" 
        name="websocket-api" 
        context="/chats/1.0.0" 
        version="1.0.0" 
        version-type="context"
        binds-to="WebSocketInboundEndpoint, WebSocketInboundEndpoint2"
>
    <resource uri-template="/rooms/{roomID}" faultSequence="fault">
        ...
    </resource>
    <resource uri-template="/lobby" faultSequence="fault">
        ...
    </resource>
</api>
```
* Since both the Resources don't have `binds-to`, each of them will inherit `binds-to` from the API level.
* Therefore, we can invoke `/chats/1.0.0/rooms/{roomID}` and `/chats/1.0.0/lobby` via either `WebSocketInboundEndpoint` or `WebSocketInboundEndpoint2`.

**Example 2: `binds-to` specified at the API level and Resource level**
```xml
<api xmlns="http://ws.apache.org/ns/synapse" 
        name="websocket-api" 
        context="/chats/1.0.0" 
        version="1.0.0" 
        version-type="context"
        binds-to="WebSocketInboundEndpoint, WebSocketInboundEndpoint2"
>
    <resource uri-template="/rooms/{roomID}" faultSequence="fault">
        ...
    </resource>
    <resource uri-template="/lobby" faultSequence="fault" binds-to="WebSocketInboundEndpoint">
        ...
    </resource>
</api>
```
* Since the first Resource doesn't have `binds-to`, it will inherit `binds-to` from the API level. This will behave as same as **Example 1**.
* Second Resource has a defined `binds-to`. Therefore, we can only invoke `/chats/1.0.0/lobby` from `WebSocketInboundEndpoint`.

**Example 3: Default binding**
```xml
<api xmlns="http://ws.apache.org/ns/synapse" 
        name="websocket-api" 
        context="/chats/1.0.0" 
        version="1.0.0" 
        version-type="context"
>
<resource uri-template="/rooms/{roomID}" faultSequence="fault">
        ...
    </resource>
    <resource uri-template="/lobby" faultSequence="fault">
        ...
    </resource>
</api>
```
* This doesn't have any `binds-to` specified, therefore, this will be bound to `default` (Specifying `binds-to="default"` at the API level gives the same results as well).
* Both `chats/1.0.0/rooms/{roomID}` and `chats/1.0.0/lobby`, can be invoked from any inbound endpoint.

### Refactoring: Moving classes from `org.apache.synapse.rest` to `org.apache.synapse.api`
In addition to the existing `RestRequestHandler`, this PR will introduce the `InboundApiHandler` which handles inbound API dispatching.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/api/API.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/API.java
@@ -16,7 +16,7 @@
 * under the License.
 */
 
-package org.apache.synapse.rest;
+package org.apache.synapse.api;
 
 import org.apache.axis2.Constants;
 import org.apache.commons.logging.Log;
@@ -37,11 +37,13 @@ import org.apache.synapse.commons.CorrelationConstants;
 import org.apache.synapse.config.xml.rest.VersionStrategyFactory;
 import org.apache.synapse.core.SynapseEnvironment;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
-import org.apache.synapse.rest.dispatch.DispatcherHelper;
-import org.apache.synapse.rest.dispatch.RESTDispatcher;
-import org.apache.synapse.rest.version.DefaultStrategy;
-import org.apache.synapse.rest.version.URLBasedVersionStrategy;
-import org.apache.synapse.rest.version.VersionStrategy;
+import org.apache.synapse.api.dispatch.DispatcherHelper;
+import org.apache.synapse.api.dispatch.RESTDispatcher;
+import org.apache.synapse.api.version.DefaultStrategy;
+import org.apache.synapse.api.version.URLBasedVersionStrategy;
+import org.apache.synapse.api.version.VersionStrategy;
+import org.apache.synapse.rest.Handler;
+import org.apache.synapse.rest.RESTConstants;
 import org.apache.synapse.transport.customlogsetter.CustomLogSetter;
 import org.apache.synapse.transport.http.conn.SynapseDebugInfoHolder;
 import org.apache.synapse.transport.http.conn.SynapseWireLogHolder;
@@ -51,13 +53,15 @@ import org.apache.synapse.transport.passthru.config.PassThroughConfiguration;
 import org.apache.synapse.util.logging.LoggingUtils;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-public class API extends AbstractRESTProcessor implements ManagedLifecycle, AspectConfigurable, SynapseArtifact {
+public class API extends AbstractRequestProcessor implements ManagedLifecycle, AspectConfigurable, SynapseArtifact {
 
     private String host;
     private int port = -1;
@@ -76,6 +80,8 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
     private VersionStrategy versionStrategy = new DefaultStrategy(this);
 
     private String fileName;
+
+    private Set<String> bindsTo = new HashSet<>();
 
     private  Log apiLog;
     private static final Log trace = LogFactory.getLog(SynapseConstants.TRACE_LOGGER);
@@ -102,7 +108,7 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
         if (!context.startsWith("/")) {
             handleException("API context must begin with '/' character");
         }
-        this.context = RESTUtils.trimTrailingSlashes(context);
+        this.context = ApiUtils.trimTrailingSlashes(context);
         apiLog = LogFactory.getLog(SynapseConstants.API_LOGGER_PREFIX + name);
 
     }
@@ -247,7 +253,19 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
         return handlers.toArray(new Handler[handlers.size()]);
     }
 
-    boolean canProcess(MessageContext synCtx) {
+    public Set<String> getBindsTo() {
+        return bindsTo;
+    }
+
+    public void addBindsTo(String inboundEndpointName) {
+        bindsTo.add(inboundEndpointName);
+    }
+
+    public void addAllBindsTo(Set<String> inboundEndpointBindings) {
+        this.bindsTo.addAll(inboundEndpointBindings);
+    }
+
+    public boolean canProcess(MessageContext synCtx) {
         if (synCtx.isResponse()) {
             String apiName = (String) synCtx.getProperty(RESTConstants.SYNAPSE_REST_API);
             String version = synCtx.getProperty(RESTConstants.SYNAPSE_REST_API_VERSION) == null ?
@@ -257,9 +275,9 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
                 return false;
             }
         } else {
-            String path = RESTUtils.getFullRequestPath(synCtx);
+            String path = ApiUtils.getFullRequestPath(synCtx);
             if (null == synCtx.getProperty(RESTConstants.IS_PROMETHEUS_ENGAGED) &&
-                    (!RESTUtils.matchApiPath(path, context))) {
+                    (!ApiUtils.matchApiPath(path, context))) {
                 auditDebug("API context: " + context + " does not match request URI: " + path);
                 return false;
             }
@@ -317,7 +335,7 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
         return true;
     }
 
-    void process(MessageContext synCtx) {
+    public void process(MessageContext synCtx) {
 
         auditDebug("Processing message with ID: " + synCtx.getMessageID() + " through the " +
                     "API: " + name);
@@ -391,7 +409,7 @@ public class API extends AbstractRESTProcessor implements ManagedLifecycle, Aspe
             return;
         }
 
-        String path = RESTUtils.getFullRequestPath(synCtx);
+        String path = ApiUtils.getFullRequestPath(synCtx);
         String subPath;
         if (versionStrategy.getVersionType().equals(VersionStrategyFactory.TYPE_URL)) {
             //for URL based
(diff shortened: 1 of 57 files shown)


modules/core/src/main/java/org/apache/synapse/api/AbstractRequestProcessor.java
```diff
@@ -28,13 +28,13 @@
  * first invoke the canProcess method of the processor to validate whether this processor
  * can process the given request or not.
  */
-public abstract class AbstractRESTProcessor {
+public abstract class AbstractRequestProcessor { // TODO rename to AbstractRequestProcessor and move
```
Nirothipan, line 31: we have a todo here

modules/core/src/main/java/org/apache/synapse/config/xml/rest/APIFactory.java
```diff
@@ -157,6 +165,15 @@ public static API createAPI(OMElement apiElt, Properties properties) {
         return api;
     }
 
+    private static void validateBindsTo(API api, Resource resource) {
```
Nirothipan, line 168: let's rename this method. This is not only validating, but also adding bindings

modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -77,6 +82,8 @@
 
     private String fileName;
 
+    private Set<String> inboundEndpointBindings = new HashSet<>();
```
Nirothipan, line 85: shall we make this as just bindsTo ?

modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -257,7 +276,7 @@ boolean canProcess(MessageContext synCtx) {
                 return false;
             }
         } else {
-            String path = RESTUtils.getFullRequestPath(synCtx);
+            String path = ApiUtils.getFullRequestPath(synCtx);
             if (null == synCtx.getProperty(RESTConstants.IS_PROMETHEUS_ENGAGED) &&
                     (!RESTUtils.matchApiPath(path, context))) {
```
Nirothipan, line 281: matchApiPath is deprecated, please update this one. Also please check none of the deprecated ones are used in synapse itself or anywhere in wso2 code

modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -477,6 +496,15 @@ void process(MessageContext synCtx) {
         }
     }
 
+    private boolean isBound(Resource resource, MessageContext synCtx) {
```
Nirothipan, line 499: shall we add a small comment on what this method is doing

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        return url;
+    }
+
+    public static String getFullRequestPath(MessageContext synCtx) {
+        Object obj = synCtx.getProperty(RESTConstants.REST_FULL_REQUEST_PATH);
+        if (obj != null) {
+            return (String) obj;
+        }
+
```
Nirothipan, line 79: shall we remove these empty lines

modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -477,6 +496,15 @@ void process(MessageContext synCtx) {
...
+        Collection<String> bindings = resource.getInboundEndpointBindings();
```
Nirothipan, line 500: please check whether de serialization shows these attributes in ui

modules/core/src/main/java/org/apache/synapse/api/Resource.java
```diff
@@ -64,6 +66,8 @@
 
     AspectConfiguration aspectConfiguration;
 
+    private Set<String> inboundEndpointBindings = new HashSet<>();
```
Nirothipan, line 69: > bindsTo

modules/core/src/main/java/org/apache/synapse/config/xml/rest/APIFactory.java
```diff
@@ -86,12 +88,18 @@ public static API createAPI(OMElement apiElt, Properties properties) {
...
+
         Iterator resources = apiElt.getChildrenWithName(new QName(
                 XMLConfigConstants.SYNAPSE_NAMESPACE, "resource"));
         boolean noResources = true;
         while (resources.hasNext()) {
             OMElement resourceElt = (OMElement) resources.next();
-            api.addResource(ResourceFactory.createResource(resourceElt, properties));
+            Resource resource = ResourceFactory.createResource(resourceElt, properties);
+            if (!resource.getInboundEndpointBindings().isEmpty()) {
+                InboundApiUtils.validateBindsTo(api, resource);
```
Nirothipan, line 100: this seems to be done already

modules/core/src/main/java/org/apache/synapse/api/Resource.java
```diff
@@ -213,6 +217,18 @@ public void setProtocol(int protocol) {
         this.protocol = protocol;
     }
 
+    public Set<String> getInboundEndpointBindings() {
+        return inboundEndpointBindings;
+    }
+
+    public void addInboundEndpointBinding(String inboundEndpointName) {
```
Nirothipan, line 224: this method is not used anywhere, shall we remove

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
```diff
@@ -525,6 +551,18 @@ public synchronized void removeAPI(String name) {
         }
     }
 
+    private void removeBindsToMappings(String apiName) {
+        Iterator<Map.Entry<String, Map<String, API>>> iterator = apiTableWithBindsTo.entrySet().iterator();
+        while (iterator.hasNext()) {
+            Map.Entry<String, Map<String, API>> mappings = iterator.next();
+            Map<String, API> apis = mappings.getValue();
+            apis.remove(apiName);
```
chanikag, line 559: Can apis null here?

modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXMLConfigurationSerializer.java
```diff
@@ -1,288 +1,288 @@
-/*
```
chanikag, line 1: This file has changed completely by this PR. Shall we revert that?

modules/core/src/main/java/org/apache/synapse/config/SynapseConfiguration.java
senthuran16: No, since we don't put `null` valued `API`s to this map.

modules/core/src/main/java/org/apache/synapse/config/xml/SynapseXMLConfigurationSerializer.java
senthuran16: Only an import was modified in this class:
```diff
-import org.apache.synapse.rest.API;
+import org.apache.synapse.api.API;
```
Even though the complete file appears as a diff here, this line is shown as the only diff in the IDE.

modules/core/src/main/java/org/apache/synapse/api/API.java
senthuran16: Checked and confirmed that these attributes were not shown, resolved it with 6512d31.

modules/core/src/main/java/org/apache/synapse/api/API.java
senthuran16: Addressed in d24336b

modules/core/src/main/java/org/apache/synapse/api/AbstractRequestProcessor.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/api/API.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/api/API.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/config/xml/rest/APIFactory.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/api/Resource.java
senthuran16: Addressed in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/config/xml/rest/APIFactory.java
senthuran16: This method is not used anymore, it has been moved to `ApiUtils`, where only validation is done. Removed this in 89315e2 and cc02c4c

modules/core/src/main/java/org/apache/synapse/api/Resource.java
senthuran16: Addressed in 89315e2 and cc02c4c

Nirothipan (approved):

PR #1742, opened 2021-03-11
remove correlation enable check from correlation id peristance

tharindu1st:
Fix https://github.com/wso2/product-apim/issues/10237


modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -371,12 +371,10 @@ public void process(MessageContext synCtx) {
 		}
         if (synCtx.isResponse()) {
             org.apache.axis2.context.MessageContext context = ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-            if (context.isPropertyTrue(PassThroughConstants.CORRELATION_LOG_STATE_PROPERTY)) {
```
Nirothipan, line 374: seems we have removed all the places where we check for this property, shall we remove the setters too, since its of no use.
For eg: https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java#L389

Nirothipan (approved):

PR #1745, opened 2021-03-15
Fix escaping of JSON objects and arrays

SanojPunchihewa:
## Purpose
Fixes https://github.com/wso2/micro-integrator/issues/2086

## User stories
With this fix, if you want to evaluate a valid JSON object as a string, you need to use **literal="true"** in the PayloadFactoryMediator as given below, 

```xml
<payloadFactory media-type="json">
<format> { "message":{ "payload": "$1" } } 
</format>
<args>
<arg evaluator="xml" expression="$ctx:jsonPayload" literal="true" />
</args>
</payloadFactory> 
```


modules/core/src/test/java/org/apache/synapse/mediators/transform/pfutils/TemplateProcessorTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private static ArgumentDetails getArgumentDetails(String pathType, boolean isXML, boolean isLiteral) {
+
+        ArgumentDetails argumentDetails = new ArgumentDetails();
+        argumentDetails.setPathType(pathType);
+        argumentDetails.setXml(isXML);
+        argumentDetails.setLiteral(isLiteral);
+        return argumentDetails;
+    }
+}
```
Nirothipan, line 218: let's add a new line here

SanojPunchihewa: Fixed with d96163f

Nirothipan (approved):

abeykoon:
@SanojPunchihewa Make sure this information goes to the documentation.

PR #1766, opened 2021-04-01
Fix issues when response with an unsupported content type received 

chashikajw:
### Purpose
This PR fixes the issues below.

1. OOM issue occurs when the response is containing an unsupported content type
2. Client does not get any response if there is any exception occurred during the execution of send mediator.

### Issues
Fixes https://github.com/wso2/product-apim/issues/9672


isudana (approved):

isudana:
It looks like we need to review this fix again more carefully. I am closing this PR and will be tracking this through 
https://github.com/wso2/product-apim/issues/9672

PR #1772, opened 2021-04-06
fix performance issue

tharindu1st:
fix performance issue

When dispatchToAPI method get hit from request and response  at the response we can use object already available in response message context PROCESSED_API

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
@@ -26,8 +26,11 @@ import org.apache.synapse.api.API;
 import org.apache.synapse.api.AbstractApiHandler;
 import org.apache.synapse.api.ApiConstants;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
+import org.apache.synapse.rest.RESTConstants;
 
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 
 /**
  * This class is responsible for receiving requests from various sources and dispatching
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
```diff
@@ -69,8 +72,14 @@ public boolean process(MessageContext synCtx) {
...
-        Collection<API> apis = synCtx.getEnvironment().getSynapseConfiguration().getAPIs(
-                ApiConstants.DEFAULT_BINDING_ENDPOINT_NAME);
+        Object apiObject = synCtx.getProperty(RESTConstants.PROCESSED_API);
+        Collection<API> apis;
+        if (apiObject != null) {
+            apis = Collections.singletonList((API) apiObject);
+        } else {
+            apis = synCtx.getEnvironment().getSynapseConfiguration().getAPIs(
+                    ApiConstants.DEFAULT_BINDING_ENDPOINT_NAME);
+        }
```
Nirothipan, line 82: isn't this anyway done in abstract class, here https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/api/AbstractApiHandler.java#L52 when called in line 84

tharindu1st: RestRequestHandler hits first before the AbstractAPIHandler Either We handle the code at AbstractAPIHandler, code get obselete due to 
           apis = synCtx.getEnvironment().getSynapseConfiguration().getAPIs(
                    ApiConstants.DEFAULT_BINDING_ENDPOINT_NAME);
Lock the SynapseConfiguration Object.

isudana (approved):

PR #1896, opened 2022-01-18
Fix Content-Type wrongly set when using SWA

malakaganga:
For the SWA, content-type is set to text/xml but this need to be
multipart/related. Fix this by changing targetRequestFactory

Fixes: wso2/product-apim#12104

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
@@ -188,11 +188,11 @@ public class TargetRequest {
 		if (o != null && o instanceof TreeMap) {
 			Map _headers = (Map) o;
 			String trpContentType = (String) _headers.get(HTTP.CONTENT_TYPE);
-			if (trpContentType != null && !trpContentType.equals("")) {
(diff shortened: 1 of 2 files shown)


sachithKay (approved):

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/TargetRequest.java
```diff
@@ -188,11 +188,11 @@ public void start(NHttpClientConnection conn) throws IOException, HttpException
 		if (o != null && o instanceof TreeMap) {
 			Map _headers = (Map) o;
 			String trpContentType = (String) _headers.get(HTTP.CONTENT_TYPE);
-			if (trpContentType != null && !trpContentType.equals("")) {
-				if (!TargetRequestFactory.isMultipartContent(trpContentType)) {
-					addHeader(HTTP.CONTENT_TYPE, trpContentType);
-				}
-			}
+            if (trpContentType != null && !trpContentType.equals("")) {
```
sachithKay, line 191: formatting seems to be off

malakaganga: This is because in original file it has used tabs as spacing.

PR #1900, opened 2022-02-07
HTTP Transport on Netty

sajinieKavindya:
## Purpose
By design, the WebSocket handshake starts with an HTTP request/response, allowing servers to handle HTTP connections as well as WebSocket connections on the same port. However, in Synapse runtime, the WebSocket Transport and the Pass-Through Transport (default HTTP transport) operate on different ports. In fact, they cannot run on the same port. Due to this reason, even though a backend has its WebSocket and HTTP endpoints exposed over the same port, we still need two different endpoint URLs with two different ports to connect to its WebSocket and HTTP endpoints. The recent requirement is the GraphQL use case, where the WSO2 API-M is used to manage GraphQL backend that has WebSocket and HTTP endpoints exposed on the same port.

**Current limitation**
The limitation to using two different ports for HTTP and WebSocket connections comes with the current architecture of the product’s runtime. That is, the runtime currently has the Pass-Through Transport (PTT) implemented on Apache HTTPCore-NIO and the WebSocket Transport implemented on the Netty framework. Therefore, these two modes of transport run on two different server processes and hence, operate on different ports.

**Proposed Solution**
If both the HTTP and WebSocket transports are implemented on the same underlying framework, a single server process should be able to handle both the HTTP and WebSocket connections over a single port. Therefore, as the solution, the Pass-Through Transport (PTT) will be implemented on the Netty framework so that both HTTP and WebSocket transport can rely on the same underlying framework.

**Implementation**
Only the HTTP transport support based on Netty is added with this PR. This Netty-based HTTP transport will be started using an Inbound Endpoint in order to give the ability to create inbound messaging channels dynamically that allow HTTP traffic. 

Support for WebSocket on the same port will be added later.

Also, this adds the MessagingHandler support to the inbound endpoint.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
@@ -27,8 +27,8 @@ import org.apache.synapse.api.AbstractApiHandler;
 import org.apache.synapse.api.ApiConstants;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.rest.RESTConstants;
+import org.apache.synapse.transport.netty.BridgeConstants;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 
@@ -60,7 +60,9 @@ public class RestRequestHandler extends AbstractApiHandler {
         org.apache.axis2.context.MessageContext msgCtx = ((Axis2MessageContext) synCtx).
                 getAxis2MessageContext();
         String protocol = msgCtx.getIncomingTransportName();
-        if (!Constants.TRANSPORT_HTTP.equals(protocol) && !Constants.TRANSPORT_HTTPS.equals(protocol)) {
+        if (!Constants.TRANSPORT_HTTP.equals(protocol) && !Constants.TRANSPORT_HTTPS.equals(protocol)
+                && !BridgeConstants.TRANSPORT_HTTPWS.equals(protocol)
+                && !BridgeConstants.TRANSPORT_HTTPSWSS.equals(protocol)) {
             if (log.isDebugEnabled()) {
                 log.debug("Invalid protocol for REST API mediation: " + protocol);
             }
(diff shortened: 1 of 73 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/SynapseCallbackReceiver.java
```diff
@@ -579,16 +581,16 @@ private void handleMessage(String messageID ,MessageContext response,
...
-
-                OAuthConfiguredHTTPEndpoint httpEndpoint = (OAuthConfiguredHTTPEndpoint) successfulEndpoint;
-
-                if (originalMC != null && OAuthUtils.retryOnOAuthFailure(httpEndpoint, synapseInMessageContext,
-                        synapseOutMsgCtx)) {
-                    httpEndpoint.retryCallWithNewToken(originalMC);
-                    return;
-                }
-            }
+//            if (successfulEndpoint instanceof OAuthConfiguredHTTPEndpoint) {
```
arunans23, line 584: Why are we commenting out this part?

modules/core/src/main/java/org/apache/synapse/api/rest/RestRequestHandler.java
```diff
@@ -60,7 +60,9 @@ public boolean process(MessageContext synCtx) {
         org.apache.axis2.context.MessageContext msgCtx = ((Axis2MessageContext) synCtx).
                 getAxis2MessageContext();
         String protocol = msgCtx.getIncomingTransportName();
-        if (!Constants.TRANSPORT_HTTP.equals(protocol) && !Constants.TRANSPORT_HTTPS.equals(protocol)) {
+        if (!Constants.TRANSPORT_HTTP.equals(protocol) && !Constants.TRANSPORT_HTTPS.equals(protocol)
+                && !BridgeConstants.TRANSPORT_HTTPWS.equals(protocol)
+                && !BridgeConstants.TRANSPORT_HTTPSWSS.equals(protocol)) {
```
Nirothipan, line 65: shall we have these 4 in a predefined list and check if it doesn't contain and do this..

sajinieKavindya: Fixed in [6aa6388](https://github.com/wso2/wso2-synapse/pull/1900/commits/6aa6388510737962df81431d8c85c1463e0ed4f5)

tharindu1st (approved):

Nirothipan (approved):

arunans23 (approved):

PR #1943, opened 2022-05-04
Enable HTTP/2 in Synapse Netty transport

thuva9872:
## Purpose
Enable HTTP/2 in Synapse Netty transport.

## Goals
Enable HTTP/2 request-response support.

## Approach
The Netty transport is implemented in the Synapse using the [transport-http](https://github.com/wso2/transport-http) repository. It supports both HTTP/1.1 and HTTP/2. Some modifications were done to the existing Netty HTTP/1.1 transport implementation in Synapse to support HTTP/2.

## Test environment
JDK 1.8
OS: Ubuntu 20.04

### DIFF
--- a/modules/commons/pom.xml
+++ b/modules/commons/pom.xml
@@ -23,7 +23,7 @@
     <parent>
         <groupId>org.apache.synapse</groupId>
         <artifactId>Apache-Synapse</artifactId>
-        <version>2.1.7-wso2v273</version>
+        <version>2.1.7-wso2v274-SNAPSHOT</version>
         <relativePath>../../pom.xml</relativePath>
     </parent>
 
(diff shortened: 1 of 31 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/netty/BridgeConstants.java
```diff
@@ -109,6 +110,7 @@ public class BridgeConstants {
 
     public static final String PROTOCOL_HTTP = "http";
     public static final String PROTOCOL_HTTPS = "https";
+    public static final String CHANNEL_HANDLER_CONTEXT = "CHNL_HNDLR_CTX";
```
Nirothipan, line 113: shall we use the full name for variable value also

PR #1949, opened 2022-05-09
Enable HTTP/2 in Netty Tranport

thuva9872:
## Purpose
Enable HTTP/2 in Synapse Netty transport.

## Goals
Enable HTTP/2 request-response support.

## Approach
The Netty transport is implemented in the Synapse using the [transport-http](https://github.com/wso2/transport-http) repository. It supports both HTTP/1.1 and HTTP/2. Some modifications were done to the existing Netty HTTP/1.1 transport implementation in Synapse to support HTTP/2.

## Test environment
JDK 1.8
OS: Ubuntu 20.04


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/netty/listener/Axis2HttpTransportListener.java
```diff
@@ -79,6 +79,13 @@ public void init(ConfigurationContext configurationContext, TransportInDescripti
     public void start() throws AxisFault {
         ServerConnectorFuture serverConnectorFuture = serverConnector.start();
         serverConnectorFuture.setHttpConnectorListener(new Axis2HttpConnectorListener(sourceConfiguration));
+        if (LOG.isDebugEnabled()) {
+            if (sourceConfiguration.getProtocol().equals("2.0")) {
```
Nirothipan, line 83: can we use the variable protocol here

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/netty/listener/HttpRequestWorker.java
```diff
@@ -338,7 +338,12 @@ private void sendResponse(int statusCode, HttpResponseStatus responseStatus, boo
 
         HttpCarbonMessage outboundResponse;
         try {
-            outboundResponse = new HttpCarbonMessage(new DefaultHttpResponse(HttpVersion.HTTP_1_1, responseStatus));
+            HttpVersion version = HttpVersion.HTTP_1_1;
+            if (sourceConfiguration.getProtocol().equals(BridgeConstants.HTTP_2_0_VERSION)) {
```
sajinieKavindya, line 342: Shall we call the equals() method on BridgeConstants.HTTP_2_0_VERSION instead of on sourceConfiguration.getProtocol() to avoid any NPEs?

thuva9872: changed it

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/netty/sender/SourceResponseHandler.java
```diff
@@ -285,6 +289,8 @@ private static String determineHttpVersion(MessageContext msgContext) {
 
         if (msgContext.isPropertyTrue(BridgeConstants.FORCE_HTTP_1_0)) {
             return "1.0";
+        } else if (BridgeConstants.HTTP_2_0_VERSION.equals(((HttpCarbonMessage) msgContext.getProperty(BridgeConstants.HTTP_CARBON_MESSAGE)).getHttpVersion())) {
+            return "2.0";
```
Nirothipan, line 293: let's use constant here too

Nirothipan (dismissed):

Nirothipan (approved):

PR #1965, opened 2022-06-20
Add changes required to build using JDK 11

rosensilva:
Related issue: https://github.com/wso2/api-manager/issues/179


modules/commons/src/main/java/org/apache/synapse/commons/staxon/core/event/SimpleXMLEventAllocator.java
```diff
@@ -378,8 +378,10 @@ public QName getName() {
         }
 
         @Override
-        public Iterator<?> getNamespaces() {
-            return namespaces.iterator();
+        public Iterator<Namespace> getNamespaces() {
+            @SuppressWarnings("unchecked")
```
isudana, line 382: What's the reason for SuppressWarnings?

rosensilva: This is due to the casting of a List type. The casting is Upcast. 
```List<Namespace> namespaceList = (List<Namespace>)(List<?>) namespaces;```
Otherwise need to iterate through the whole list and copy it. Or change a considerable amount of Method Signatures and their usage. Btw, do we have an easy way to get rid of this warning?

File: modules/extensions/src/main/java/org/apache/synapse/mediators/bsf/ScriptMediator.java [hunk 2]
```diff
@@ -364,12 +355,15 @@ private ScriptMessageContext getScriptMessageContext(MessageContext synCtx, XMLH
         ScriptMessageContext scriptMC;
         if (language.equals(NASHORN_JAVA_SCRIPT)) {
             try {
-                emptyJsonObject = (ScriptObjectMirror) scriptEngine.eval("({})");
-                jsonSerializer = (ScriptObjectMirror) scriptEngine.eval("JSON");
+                if(isJDKContainNashorn()) {
```
isudana, line 358: Looks like we are calling this for all the messages, can't we do it at the initialization?

rosensilva: yeah, we should. missed this while refactoring. Now the isJDKContainNashorn method gets the value from the initialized value

isudana (approved):

PR #1984, opened 2022-07-18
Added OtlpExporter with OpenTelemetry Support

Rashmika998:
```
[opentelemetry]
enable = true
logs = true
type = “otlp”
url = “endpoint-url”

[[opentelemetry.properties]]
name = “name-of-the-header”
value = “key-value-of-the-header” 
```
if anyone uses NewRelic APM to view the traces, the following configurations should be added.
```
[opentelemetry]
enable = true
logs = true
type = “otlp”
url = “https://otlp.nr-data.net:4317/v1/traces”

[[opentelemetry.properties]]
name = “api-key”
value = “<your-insight-insert-key>” 
```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
+++ b/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
@@ -53,7 +53,7 @@ public class OTLPTelemetryManager implements OpenTelemetryManager {
 
         String headerProperty = getHeaderKeyProperty();
         if (headerProperty == null) {
-            throw new SynapseException("No properties found starting with opentelemetry.otlp.header_properties");
+            throw new SynapseException("No properties found starting with opentelemetry.properties");
         }
         String headerKey = headerProperty.substring(TelemetryConstants.OPENTELEMETRY_PROPERTIES_PREFIX.length());
         String endPointURL = SynapsePropertiesLoader.getPropertyValue(TelemetryConstants.OPENTELEMETRY_URL, null);
--- a/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
+++ b/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
@@ -1,3 +1,21 @@
+/*
+ *  Copyright (c) 2022, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
```diff
@@ -0,0 +1,106 @@
+package org.apache.synapse.aspects.flow.statistics.tracing.opentelemetry.management;
```
senthuran16, line 1: Let's add a license header

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
```diff
@@ -0,0 +1,106 @@
...
+    private Log logger = LogFactory.getLog(OTLPTelemetryManager.class);
+    private SdkTracerProvider sdkTracerProvider;
+    private OpenTelemetry openTelemetry;
+    private TelemetryTracer tracer;
+    private SpanHandler handler;
+
+    @Override
+    public void init() {
+
+        String headerProperty = getHeaderKeyProperty();
```
senthuran16, line 35: `getHeaderKeyProperty()` can return `null`. Shall we handle that, if `headerProperty == null` we throw an error saying that the required property is not found?

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/TelemetryConstants.java
```diff
@@ -55,6 +55,7 @@ public class TelemetryConstants {
     public static final String DEFAULT_ZIPKIN_HOST = "localhost";
     public static final String DEFAULT_ZIPKIN_PORT = "9411";
     public static final String ZIPKIN_API_CONTEXT = "/api/v2/spans";
+    public static final String OPENTELEMETRY_PROPERTIES = "opentelemetry.otlp.header_properties.";
```
senthuran16, line 58: Shall we make this `OPENTELEMETRY_PROPERTIES_PREFIX`?

senthuran16 (changes requested):

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
Rashmika998: Done and resolved

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/TelemetryConstants.java
Rashmika998: Done and resolved

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/OTLPTelemetryManager.java
Rashmika998: Yeah we can but it won't be null unless the properties are not set through toml file by mistakenly as the header and end point url is required.

Rashmika998: Resolved after discussion.

tharindu1st (dismissed):

senthuran16 (approved):

PR #2013, opened 2022-10-25
Build coordination between VFS sender and Receiver

malakaganga:
## Purpose
Fixes: https://github.com/wso2/api-manager/issues/806

### DIFF
--- a/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
+++ b/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
@@ -169,7 +169,7 @@ public class VFSTransportListener extends AbstractPollingTransportListener<PollT
             fsm.setConfiguration(getClass().getClassLoader().getResource("providers.xml"));
             fsm.init();
             this.workerPool = super.workerPool;
-            fsManager = fsm;
+            setFsManager(fsm);
             Parameter lockFlagParam = getTransportInDescription().getParameter(VFSConstants.TRANSPORT_FILE_LOCKING);
             if (lockFlagParam != null) {
                 String strLockingFlag = lockFlagParam.getValue().toString();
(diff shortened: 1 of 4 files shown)


modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -469,10 +474,15 @@ && acquireLock(fsManager, child, entry, fso, true)))
...
-                                        // tell moveOrDeleteAfterProcessing() file was success
-                                        entry.setLastPollState(PollTableEntry.SUCCSESSFUL);
+                                        boolean status = processFile(entry, child);
+                                        if (status) {
+                                            successCount++;
+                                            entry.setLastPollState(PollTableEntry.SUCCSESSFUL);
+                                        } else {
+                                            entry.setLastPollState(PollTableEntry.FAILED);
+                                        }
+
```
rosensilva, line 484: extra new line?

rosensilva (approved):

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -881,6 +890,12 @@ private void processFile(PollTableEntry entry, FileObject file) throws AxisFault
                     null, //* SOAP Action - not applicable *//
                     contentType
                 );
+                //Check if the sender was successful or not
+                String errorCode = (String) transportHeaders.get(VFSConstants.ERROR_CODE);
```
nuwanw, line 894: Can we check a synapse level parameter rather than transport scope, if this is transport, this will always a custom property that we have to set in fault sequence

malakaganga: Issue here is Existing code creates axis2 context in the VFSTransportListner and injects it to the axis.engine within AbstractTransportListner
And from that method what we receive is also an axis2 message context so we don't have the access to synapse context in this transport layer

PR #2033, opened 2022-11-30
Fix copying jms params to synapse messsage context via in proxy listener

lasanthaS:
## Purpose
This PR copies Proxy listener parameters into the Synapse Message Context.

Fixes https://github.com/wso2/api-manager/issues/936

## Approach
This PR introduces the following parameter in the Proxy listener to copy the parameters into the Synapse Message Context.

```xml
<parameter name="SET_JMS_SVC_PARAMS_AS_MSG_CTX">true</parameter>
```

## Security checks


senthuran16 (dismissed):

malakaganga (changes requested):
Shall we add the public git issue to the commit itself and also to the pr description as well rather than an internal issue?
Also this is public fix when adding public git issue to the commit message shall we add in the format

**Fixes: public_git_issue** so it will close the public issue upon pr merge

malakaganga (approved):

PR #2058, opened 2023-03-08
Fixed ftl file reference issue

chamikasudusinghe:
The reason for the issue was the expiring time of the template which was set to 1500ms. Due to that when accessing ftl via a proxy service after the API call, if the expiry time is not reached, the payload was not used.

Public Git Issue: https://github.com/wso2/api-manager/issues/1520

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
@@ -189,7 +189,8 @@ public class PayloadFactoryMediator extends AbstractMediator {
             } else if (entry instanceof String) {
                 text = (String) entry;
             }
-            if (reCreate) {
+            boolean templateLoaded = templateProcessor.getTemplateStatus();
+            if (reCreate || !templateLoaded) {
                 templateProcessor.setFormat(text);
                 templateProcessor.init();
             }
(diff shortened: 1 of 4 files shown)


modules/core/src/test/java/org/apache/synapse/mediators/transform/pfutils/FreeMarkerTemplateProcessorTest.java
```diff
@@ -421,6 +422,54 @@ public void testWithXMLPayloadWithFtlHeader() throws Exception {
                 + "set expected format", expectedEnvelopeFtl, synCtx.getEnvelope().getBody().toString());
     }
 
+    public void testMultipleFtlPayloads() throws Exception {
```
SanojPunchihewa, line 425: We need to have an integration test to validate this fix. Since you have explicitly called `init()` in the two templateProcessor instances, this will work even without the fix. 

We need to write a synapse sequence with two Payload factory mediators and invoke it and verify. Please refer https://github.com/wso2/wso2-synapse/pull/1739/files to get an idea. If you need further clarifications let's have an offline discussion.

chamikasudusinghe: Moving this test case to micro integrator from synapse

SanojPunchihewa (approved):

PR #2067, opened 2023-04-19
Close target connection upon endpoint timeout

SanojPunchihewa:
## Purpose
In the current architecture, when the endpoint timeouts the timeout handler will call the fault sequence and finish the mediation flow for that request. But the connection to the backend is not closed until the socket timeout.

This fix will improve the architecture to close the socket connection when an endpoint timeout happens for HTTP transport.

Fixes: https://github.com/wso2/api-manager/issues/1746

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/TimeoutHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/TimeoutHandler.java
@@ -21,6 +21,7 @@ package org.apache.synapse.core.axis2;
 
 import org.apache.axiom.om.OMAbstractFactory;
 import org.apache.axiom.soap.SOAPEnvelope;
+import org.apache.axis2.description.TransportOutDescription;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.FaultHandler;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/core/axis2/TimeoutHandler.java
```diff
@@ -233,6 +233,8 @@ private void processCallbacks() {
                     if (RuntimeStatisticCollector.isStatisticsEnabled()) {
                         CallbackStatisticCollector.callbackCompletionEvent(callback.getSynapseOutMsgCtx(), (String) key);
                     }
+                    // Call the TransportSender's onAppError method to release any resources
+                    callback.getAxis2OutMsgCtx().getTransportOut().getSender().onAppError(callback.getAxis2OutMsgCtx());
                 }
```
isudana, line 238: Can there be a NPE? if sender is not available for some reason? shall we handle those cases?

SanojPunchihewa: The sending flow will break if the sender is not available and won't reach this. As a safety measure, I'll add a null check

SanojPunchihewa: Fixed with 0328c2c

isudana (approved):

SanojPunchihewa:
Closing this PR, since the changes are added via https://github.com/wso2/wso2-synapse/pull/2076

PR #2077, opened 2023-06-12
Fix Passthrough Threads getting stuck due to request message discard

malakaganga:
## Purpose
Adding new worker pool (RequestMessageDiscardWorkerPool) to discard request messages So ClientWorker (Passthrough Message Processor) threads won't get stucked while draining the input stream. Also remove setting REQUEST_DONE state in SourceContext when Error Scenario since when discarding the message we automatically set REQUEST_DONE in SourceHandler.

Fixes: https://github.com/wso2/api-manager/issues/1792

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/MessageDiscardWorker.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/MessageDiscardWorker.java
@@ -1,0 +1,97 @@
+/**
+ *  Copyright (c) 2023, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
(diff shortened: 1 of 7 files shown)


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/MessageDiscardWorker.java
```diff
@@ -0,0 +1,97 @@
+/**
+ *  Copyright (c) 2023, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
arunans23, line 2:
```suggestion
 *  Copyright (c) 2023, WSO2 LLC. (http://www.wso2.org)
```

arunans23 (dismissed):

arunans23 (approved):

PR #2082, opened 2023-07-04
Add aouth connection timeout for endpoint

madushadhanushka:
Fix https://github.com/wso2/api-manager/issues/1939
The following three properties added as timeout for oauth endpoint validation:
- connectionTimeout: the time to establish the connection with the remote host
- connectionRequestTimeout:  This timeout is used to determine how long the client will wait for a connection from the connection manager before giving up.
- socketTimeout: the time waiting for data – after establishing the connection; maximum time of inactivity between two data packets
```
<authentication>
<oauth>
<clientCredentials>
<clientId>8ghGvzmw3itBfNSkicxs2aWQzzoa</clientId>
<clientSecret>XXXXXXXXXXXXXXx</clientSecret>
<tokenUrl>https://localhost:9444/oauth2/token</tokenUrl>
<connectionTimeout>1000</connectionTimeout>
<connectionRequestTimeout>1000</connectionRequestTimeout>
<socketTimeout>1000</socketTimeout>
</clientCredentials>
</oauth>
</authentication>
```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/endpoints/auth/oauth/AuthorizationCodeHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/endpoints/auth/oauth/AuthorizationCodeHandler.java
@@ -34,9 +34,11 @@ public class AuthorizationCodeHandler extends OAuthHandler {
     private final String refreshToken;
 
     public AuthorizationCodeHandler(String tokenApiUrl, String clientId, String clientSecret,
-                                    String refreshToken, String authMode) {
+                                    String refreshToken, String authMode, int connectionTimeout,
+                                    int connectionRequestTimeout, int socketTimeout) {
 
-        super(tokenApiUrl, clientId, clientSecret, authMode);
+        super(tokenApiUrl, clientId, clientSecret, authMode, connectionTimeout, connectionRequestTimeout,
+                socketTimeout);
         this.refreshToken = refreshToken;
     }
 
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/synapse/endpoints/auth/oauth/AuthorizationCodeHandler.java
```diff
@@ -69,6 +71,18 @@ protected OMElement serializeSpecificOAuthConfigs(OMFactory omFactory) {
...
+                    String.valueOf(connectionTimeout)));
+        }
+        if (connectionRequestTimeout != -1) {
+            authCode.addChild(OAuthUtils.createOMElementWithValue(omFactory, AuthConstants.OAUTH_CONNECTION_REQUEST_TIMEOUT,
+                    String.valueOf(connectionRequestTimeout)));
+        }
+        if (socketTimeout != -1) {
+            authCode.addChild(OAuthUtils.createOMElementWithValue(omFactory, AuthConstants.OAUTH_SOCKET_TIMEOUT,
+                    String.valueOf(socketTimeout)));
+        }
```
SanojPunchihewa, line 85: Since these configs are common for all the OAuth types, let's move this logic to the `serializeOAuthConfiguration` [1] method in the `OAuthHandler` parent class.

[1] - https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/endpoints/auth/oauth/OAuthHandler.java#L181

madushadhanushka: Fixed

SanojPunchihewa (approved):

PR #2083, opened 2023-07-12
IdentifyApi Refactored

shnrndk:
This PR refactored the ```identifyApi()``` method. Now the api identification logic is taken from the ```canProcess()``` method to the newly created ```identifyApi()``` method. Previous ```identifiyApi()``` method was renamed to ```startProcess()``` method.
Related Issue:  https://github.com/wso2/api-manager/issues/1564

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/api/API.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/API.java
@@ -53,12 +53,12 @@ import org.apache.synapse.transport.passthru.config.PassThroughConfiguration;
 import org.apache.synapse.util.logging.LoggingUtils;
 
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.List;
 import java.util.LinkedHashMap;
 import java.util.ArrayList;
-import java.util.LinkedHashSet;
-import java.util.Collection;
 import java.util.Set;
 import java.util.HashSet;
 
@@ -276,20 +276,11 @@ public class API extends AbstractRequestProcessor implements ManagedLifecycle, A
                 return false;
             }
         } else {
-            String path = ApiUtils.getFullRequestPath(synCtx);
-            if (null == synCtx.getProperty(RESTConstants.IS_PROMETHEUS_ENGAGED) &&
-                    (!ApiUtils.matchApiPath(path, context))) {
-                auditDebug("API context: " + context + " does not match request URI: " + path);
-                return false;
+            if (ApiUtils.identifyApi(this, synCtx) == false) {
+                ApiUtils.getFullRequestPath(synCtx);
             }
-
-            if(!versionStrategy.isMatchingVersion(synCtx)){
-                return false;
-            }
-
             org.apache.axis2.context.MessageContext msgCtx =
                     ((Axis2MessageContext) synCtx).getAxis2MessageContext();
-
             if (host != null || port != -1) {
                 String hostHeader = getHostHeader(msgCtx);
                 if (hostHeader != null) {
--- a/modules/core/src/main/java/org/apache/synapse/api/API.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/API.java
@@ -280,7 +280,7 @@ public class API extends AbstractRequestProcessor implements ManagedLifecycle, A
                 return false;
             }
         } else {
-            if (ApiUtils.identifyApi(this, synCtx) == false) {
+            if (!ApiUtils.identifyApi(this, synCtx)) {
                 return false;
             }else{
                 ApiUtils.getFullRequestPath(synCtx);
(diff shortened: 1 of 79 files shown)


modules/core/src/main/java/org/apache/synapse/api/API.java
```diff
@@ -276,20 +280,14 @@ public boolean canProcess(MessageContext synCtx) {
...
-                    (!ApiUtils.matchApiPath(path, context))) {
-                auditDebug("API context: " + context + " does not match request URI: " + path);
-                return false;
-            }
-
-            if(!versionStrategy.isMatchingVersion(synCtx)){
+            if (!ApiUtils.identifyApi(this, synCtx)) {
                 return false;
+            }else{
+                ApiUtils.getFullRequestPath(synCtx);
```
isudana, line 286: can we simply move this out of the else block?

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +161,52 @@ public static String getSubRequestPath(MessageContext synCtx) {
         return (String) synCtx.getProperty(RESTConstants.REST_SUB_REQUEST_PATH);
     }
 
+    public static boolean identifyApi(API api, MessageContext synCtx) {
+        if (synCtx.getProperty(RESTConstants.IDENTIFIED_API) == null) {
+            String path = ApiUtils.getFullRequestPath(synCtx);
+            if (null == synCtx.getProperty(RESTConstants.IS_PROMETHEUS_ENGAGED) &&
```
isudana, line 167: It seems there is an issue with the original code here, what if PROMETEUS_ENGAGED is present and matchApiPath is false, we still have to return false; Let's fix that.

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +161,52 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+    /**
+     * Checks whether the provided resource is capable of processing the message from the provided message context.
+     * The resource becomes capable to do this when the it contains either the name of the api caller,
+     * or {@value ApiConstants#DEFAULT_BINDING_ENDPOINT_NAME}, in its binds-to.
+     *
+     * @param resource  Resource object
+     * @param synCtx    MessageContext object
+     * @return          Whether the provided resource is bound to the provided message context
+     */
+    public static boolean isBound(Resource resource, MessageContext synCtx) {
```
isudana, line 191: make it private

modules/core/src/main/java/org/apache/synapse/api/AbstractApiHandler.java
```diff
@@ -51,19 +51,30 @@ protected boolean dispatchToAPI(Collection<API> apiSet, MessageContext synCtx) {
...
+            if (startProcess((API) apiObject, synCtx, defaultStrategyApiSet)) {
                 return true;
             }
         } else {
             //To avoid apiSet being modified concurrently
             List<API> duplicateApiSet = new ArrayList<API>(apiSet);
-            for (API api : duplicateApiSet) {
-                if (identifyAPI(api, synCtx, defaultStrategyApiSet)) {
+            API identifiedApi = (API) synCtx.getProperty(RESTConstants.IDENTIFIED_API);
+            //If normal path this goes to else, if logging enabled then goes inside the if method.
```
isudana, line 61: This logic is not clear to me, can you please explain?

shnrndk: In the api logging flow in the first iteration it needs to identify the api. For that it calls the `identifyApi()` method. If the api is identified then that api gets stored in the IDENTIFIED_API property. Then in the second iteration, we do not need to identify the api again. Therefore we get the identified api from the from the IDENTIFIED_API property. So we do not need to iterate all the apis in the duplicateApiSet to identify the api.

But if the api logging is disabled then api is not identified previously, therefore IDENTIFIED_API property is null, So the api needs to be identified by iterating all the apis in the defaultStrategyApiSet.

modules/core/src/main/java/org/apache/synapse/api/API.java
shnrndk: Fixed

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
shnrndk: Added the new condition block, let me know if the debug log message needs to changed or not.

isudana: I still don't understand why we are checking Prometheus is enabled. Based on your logic we should be able to just get rid of the check. Shall we check the history we are checking this condition?

arunans23:
Closing since the relevant changes are added from https://github.com/wso2/wso2-synapse/pull/2150

PR #2089, opened 2023-08-08
Add transport parameters transport.vfs.MinimumAge and transport.vfs.MaximumAge

formarlyjlauper:
###Goal
This adds the transport parameters transport.vfs.MinimumAge and transport.vfs.MaximumAge which check for file age since last mod in seconds.

###Sample
`<parameter name="transport.vfs.MinimumAge">10</parameter>	
  <parameter name="transport.vfs.MaximumAge">60</parameter>`
reads files that are atleast 10 seconds old but no older than a minute.

### DIFF
--- a/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
+++ b/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
@@ -531,7 +531,6 @@ public class PollTableEntry extends AbstractPollTableEntry {
     }
 
     protected boolean loadConfigurationsFromService(ParameterInclude params) throws AxisFault {
-        System.out.println("loading config from service");
         fileURI = ParamUtils.getOptionalParam(params, VFSConstants.TRANSPORT_FILE_FILE_URI);
         if (fileURI == null) {
             log.warn("transport.vfs.FileURI parameter is missing in the proxy service configuration");
--- a/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
+++ b/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
@@ -511,10 +511,6 @@ public class PollTableEntry extends AbstractPollTableEntry {
         return maximumAge;
     }
     
-    public boolean hasAgeCheck(){
-        return minimumAge != null||maximumAge != null;
-    }
-    
     @Override
     public boolean loadConfiguration(ParameterInclude params) throws AxisFault {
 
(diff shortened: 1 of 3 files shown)


modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
```diff
@@ -515,6 +531,7 @@ public boolean loadConfiguration(ParameterInclude params) throws AxisFault {
     }
 
     protected boolean loadConfigurationsFromService(ParameterInclude params) throws AxisFault {
+        System.out.println("loading config from service");
```
arunans23, line 534: Shall we change this to use the logger? Using `System.out.println` to print logs is not encouraged.

arunans23:
Thank you for your contribution. Can you address comments in order to merge the PR?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
ljuillerat: This lines was added for debugging purpose, we just removed it

ljuillerat:
Building CI seems failing but for external reasons. Would it be possible to have a look ?

ljuillerat:
@arunans23 Based on our PR, it doesn't seem that the failing build is concerned. Would it be possible to confirm ?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
```diff
@@ -499,6 +503,18 @@ public void setSubfolderTimestamp(String subfolderTimestamp) {
...
 
+    public Long getMinimumAge() {
+        return minimumAge;
+    }
+
+    public Long getMaximumAge() {
+        return maximumAge;
+    }
+    
+    public boolean hasAgeCheck(){
```
arunans23, line 514: Are we utilizing this method?

arunans23:
> @arunans23 Based on our PR, it doesn't seem that the failing build is concerned. Would it be possible to confirm ?

Yes. It is not related to this PR. I have added a small comment. Shall we address it before merging the PR?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/PollTableEntry.java
ljuillerat: Method is effectively unused. Just removed it from this PR.

arunans23:
Thank you for addressing the comments on time. We can track the doc changes from a previous issue related to VFS. 
https://github.com/wso2/docs-mi/issues/809

arunans23 (approved):

PR #2095, opened 2023-08-24
Adding parameters to delay file processing in specific cases

ljuillerat:
## Purpose
Resolves some use cases where file transfer must be timed out. Especially where file content is not consistent or file not finalized to be retrieved. 

## Goals
2 additional properties are added to the VFS transport listener to manage:
- Empty files (boolean to state if they need to be read or not)
- Files that are not complete and where content is still changing by the time (we can define a value which is telling the check delay in ms). As long as these files are changing, they will not be picked up.

## Samples
To ignore empty files, parameter has to be added to listening proxy:
`<parameter name="transport.vfs.CheckSizeIgnoreEmpty">true</parameter>`

To check if file is still uploading, we have to add this parameter to listening proxy:
`<parameter name="transport.vfs.CheckSizeInterval">100</parameter>`


arunans23:
Thank you for your contribution. It seems this PR has some conflicts. Can you resolve them to merge the PR?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -1194,5 +1208,149 @@ public int compare(FileObject o1, FileObject o2) {
...
+
+            //wait interval time 
+            log.debug("Check if file is still uploading. Now sleep "+checkSizeInterval+" ms");
+            Thread.sleep(checkSizeInterval);
+            //get second MD5
+            //clear cache and refresh
+            fsManager.getFilesCache().close();
+            child.refresh();
+            String md5AfterSleep = getMD5Checksum(child);
+            //System.out.println("After " + checkSizeInterval + "ms second size: " + md5B);
```
arunans23, line 1254: Shall we remove this line if not needed or move to a debug log?

ljuillerat: Yes sure, the line https://github.com/wso2/wso2-synapse/pull/2095/files/dea5cb6ba6c5b8bfcf8e8b06ad7b858f2e80b19d#diff-8258f01ab96fc50493ea56c161dd25a7ca9f8835cfd76ccfab590716cb32ec1dR1254 should not be present

ljuillerat:
Building CI seems failing but for external reasons. Would it be possible to have a look ?

arunans23 (approved):

arunans23:
Thank you for addressing the review comments. Created a doc issue to track the doc changes. https://github.com/wso2/docs-mi/issues/809

PR #2098, opened 2023-09-05
Properly destroy passthru worker pools during Listener destroy stage

arunans23:
Resolves https://github.com/wso2/micro-integrator/issues/2963

HTTP and HTTPS transport listeners are sharing the Passthru worker pool threads. Hence it is wrong to shutdown the worker pool for one Listener while the other Listener is active. This fix will ensure that the worker pool threads are shutdown when the listeners are destroyed.

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpListener.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughHttpListener.java
@@ -445,7 +445,6 @@ public class PassThroughHttpListener implements TransportListener {
                 }
             }
             serviceTracker.stop();
-            handler.stop();
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


isudana (approved):

srinivaskancharla:
@arunans23 @isudana 
Could you please validate if the issue reported here https://stackoverflow.com/questions/78078997/wso2mi-4-2-0-continous-socket-connection-errors-in-logs-when-pod-count-1

is similar to the one fixed?

arunans23:
> @arunans23 @isudana Could you please validate if the issue reported here https://stackoverflow.com/questions/78078997/wso2mi-4-2-0-continous-socket-connection-errors-in-logs-when-pod-count-1
> 
> is similar to the one fixed? appreciate your response

No, this fix is done to improve the shutdown logic of passthrough transport listeners.

srinivaskancharla:
Okay, thanks for the response.

PR #2099, opened 2023-09-13
Introducing RequiredFileNamePatterns feature to read files depending on the existence of other files

olhepp-avintis:
## Documentation
Example for usage: 
```
<parameter name="transport.vfs.FileNamePattern">(.*)[.](hl7|pdf)$</parameter>
<parameter name="transport.vfs.RequiredFileNamePatterns">$1[.]pdf,$1[.]hl7</parameter>
```
> The above mentioned example shows that a file (with .hl7 or .pdf extension) is only transfered, if both regular expression defined in the “RequiredFileNamePattern” are satisfied. 
The $1 part represents the first group (parentheses) of the FileNamePattern property. 
Processing files will be processed in the order of their appearance in RequiredFileNamePatterns parameter.

### DIFF
--- a/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
+++ b/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
@@ -64,6 +64,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


arunans23:
Can you sign the CLA to merge the PR?

modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -1195,4 +1222,74 @@ public int compare(FileObject o1, FileObject o2) {
...
+				}
+        		Pattern pattern = Pattern.compile(fixedPattern);
+        		Matcher matcher = pattern.matcher(baseName);
+        		//match the "fixed" expression with named groups
+        		if(matcher.find()) {
+        			String relevantname = "";
+        			//reconstruct the filename that matters with the regex groups ($1-$9)
+        			for (String groupName: groups) {
+            			relevantname+=matcher.group(groupName);
+					}
```
arunans23, line 1287: Shall we format the code segments? It seems both tabs and spaces are used. Let's use only spaces and format the code. Please check other places as well.

arunans23:
Shall we explain in the PR description a little bit more about why this requirement came up?

chanikag:
As there are conflicts in the PR, closing it. Please reopen if required.

PR #2119, opened 2023-12-01
Do cert revocation improvement to validate with only one client certificate in the request when the issuer of that cert is present in the client trust store

Akila94:
## Purpose
- This improvement contains the certificate revocation validation improvement so the validation can be done without the full certificate chain sent in the request.

- The validation can also be done when only one intermediate client certificate is sent in the request. For this to happen the immediate issuer of that certificate should contain in the client trust store.

- Also a cert cache (LRU) is implemented to keep the issuer certificate found after traversing the client trust store.

- A data holder is introduced to avoid client trust store loading multiple times for a single server start-up.

- One of the following configurations can be used to enable this improvement,

1. Enforce glocally using axis2.xml. The following configuration should be added inside the ```<transportReceiver name="https" class="org.apache.synapse.transport.passthru.PassThroughHttpMultiSSLListener">```.

```
<parameter name="CertificateRevocationVerifier" enable="true">
    <CacheSize>1024</CacheSize>
    <CacheDelay>1000</CacheDelay>
    <FullChainValidation>false</FullChainValidation>
    <ExpiryValidation>true</ExpiryValidation>
</parameter>
```

2. Enforce per security listener profile. The following configuration should be added inside the relevant profile.

```
<CertificateRevocationVerifier>
    <Enable>true</Enable>
    <CacheSize>1024</CacheSize>
    <CacheDelay>1000</CacheDelay>
    <FullChainValidation>false</FullChainValidation>
    <ExpiryValidation>true</ExpiryValidation>
</CertificateRevocationVerifier>
```


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,263 @@
...
+                        } catch (KeyStoreException e) {
+                            throw new CertificateVerificationException("Unable to read the certificate from " +
+                                    "truststore with the alias: " + alias, e);
+                        }
+
+                        if (issuerCert == null) {
+                            throw new CertificateVerificationException("Issuer certificate not found in truststore");
+                        }
+
+                        peerCert.verify(issuerCert.getPublicKey());
```
Akila94, line 150: add try here

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,263 @@
...
+                    log.info("Path verification Successful. Took " + (System.currentTimeMillis() - start) + " ms.");
+                } else {
+                    if (log.isDebugEnabled()) {
+                        log.debug("Validating client certificate with the issuer certificate retrieved from" +
+                                "the trust store");
+                    }
+                    verifier.checkRevocationStatus(peerCert, issuerCert);
+                }
+                return;
+            } catch (Exception e) {
```
Akila94, line 204: check this

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/ServerSSLSetupHandler.java
```diff
@@ -78,7 +83,9 @@ public void verify(
 
         if (verificationManager != null) {
             try {
-                verificationManager.verifyRevocationStatus(sslsession.getPeerCertificateChain());
+                X509Certificate[] peerCertChain = sslsession.getPeerCertificateChain();
+                verificationManager.isExpired(peerCertChain);
```
Akila94, line 87: check backward compatibility

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,263 @@
...
+        return certChain;
+    }
+
+    /**
+     * Checks whether a provided certificate is expired or not at the time it is validated.
+     *
+     * @param certificates array of certificates needs to be validated
+     * @throws CertificateVerificationException if one of the certs are expired, this exception will be thrown
+     */
+    public void isExpired(javax.security.cert.X509Certificate[] certificates) throws CertificateVerificationException {
```
Akila94, line 247: return boolean

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/cache/CertCache.java
```diff
@@ -0,0 +1,183 @@
...
+        return cache;
+    }
+
+    /**
+     * This initialize the Cache with a CacheManager. If this method is called, a cache manager will not be used.
+     *
+     * @param size max size of the cache
+     * @param delay defines how frequently the CacheManager will be started
+     */
+    public void init(int size, int delay) {
```
Akila94, line 64: handle this

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/config/TrustStoreHolder.java
```diff
@@ -0,0 +1,62 @@
...
+package org.apache.synapse.transport.nhttp.config;
+
+import java.security.KeyStore;
+
+/**
+ * A data holder class to store the client trust store.
+ */
+public class TrustStoreHolder {
+
+    private static volatile TrustStoreHolder instance;
```
Akila94, line 28: remove volatile

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: This is not handled here since the next verifier should be executed if one verifier is failed. If both verifiers fail, a ```CertificateVerificationException``` is thrown in line number 209.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/nhttp/config/TrustStoreHolder.java
Akila94: The variable is made volatile to be thread-safe. Without the ```volatile``` keyword, different threads can see different states of the variable due to the Java Memory Model. Which is not a good thing.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,280 @@
...
+    private int cacheSize = Constants.CACHE_DEFAULT_ALLOCATED_SIZE;
+    private int cacheDelayMins = Constants.CACHE_DEFAULT_DELAY_MINS;
+    private boolean isFullCertChainValidationEnabled = true;
+    private boolean isCertExpiryValidationEnabled = false;
+    private static final Log log = LogFactory.getLog(CertificateVerificationManager.class);
+
+    public CertificateVerificationManager(Integer cacheAllocatedSize, Integer cacheDelayMins) {
+
+        if (cacheAllocatedSize != null && cacheAllocatedSize > Constants.CACHE_MIN_ALLOCATED_SIZE
+                && cacheAllocatedSize < Constants.CACHE_MAX_ALLOCATED_SIZE) {
```
isudana, line 61: Shall we put a warning for the out-of-range cacheAllocatedSize?

Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/be7928785b02bfa63e557a035be8fb3fa80b86f4

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,282 @@
...
+                        if (isExpired(convertedCertificates)) {
+                            throw new CertificateVerificationException("One of the provided certificates are expired");
+                        }
+                    }
+
+                    log.debug("Doing full certificate chain validation");
+                    CertificatePathValidator pathValidator = new CertificatePathValidator(convertedCertificates,
+                            verifier);
+                    pathValidator.validatePath();
+                    log.info("Path verification Successful. Took " + (System.currentTimeMillis() - start) + " ms.");
```
isudana, line 209: This info log is not required.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,282 @@
...
+                    // Unable to verify the signature.
+                    throw new CertificateVerificationException("Unable to verify the signature of the certificate.");
+                }
+            }
+        }
+
+        long start = System.currentTimeMillis();
+
+        OCSPCache ocspCache = OCSPCache.getCache();
+        ocspCache.init(cacheSize, cacheDelayMins);
```
isudana, line 188: shouldn't we initialize caches at the init, rather than doing it for each and every validation?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,282 @@
...
+                        }
+                    }
+
+                    log.debug("Validating client certificate with the issuer certificate retrieved from" +
+                            "the trust store");
+                    verifier.checkRevocationStatus(peerCert, issuerCert);
+                }
+                return;
+            } catch (Exception e) {
+                log.info(verifier.getClass().getSimpleName() + " failed.");
```
isudana, line 225: shouldn't we make this an error log?

isudana (changes requested):

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: I think this is not written as an error log since if this catch block is called, the next verifier will be hit. If we print an error log, it will give the user the wrong idea. If both verifiers fail, an exception like below is thrown at the end of the method.

```
throw new CertificateVerificationException("Path Verification Failed for both OCSP and CRL");
```

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/09b48b314fa878b08f5287ee7ab2026ac5f29a01

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: Removed with https://github.com/wso2/wso2-synapse/pull/2119/commits/09b48b314fa878b08f5287ee7ab2026ac5f29a01

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,280 @@
...
+
+        if (cacheAllocatedSize != null && cacheAllocatedSize > Constants.CACHE_MIN_ALLOCATED_SIZE
+                && cacheAllocatedSize < Constants.CACHE_MAX_ALLOCATED_SIZE) {
+            this.cacheSize = cacheAllocatedSize;
+        }
+        if (cacheDelayMins != null && cacheDelayMins > Constants.CACHE_MIN_DELAY_MINS
+                && cacheDelayMins < Constants.CACHE_MAX_DELAY_MINS) {
+            this.cacheDelayMins = cacheDelayMins;
+        }
+        this.isFullCertChainValidationEnabled = isFullCertChainValidationEnabled;
```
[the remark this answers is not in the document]
Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/09b48b314fa878b08f5287ee7ab2026ac5f29a01

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,280 @@
...
+
+    public CertificateVerificationManager(Integer cacheAllocatedSize, Integer cacheDelayMins) {
+
+        if (cacheAllocatedSize != null && cacheAllocatedSize > Constants.CACHE_MIN_ALLOCATED_SIZE
+                && cacheAllocatedSize < Constants.CACHE_MAX_ALLOCATED_SIZE) {
+            this.cacheSize = cacheAllocatedSize;
+        }
+        if (cacheDelayMins != null && cacheDelayMins > Constants.CACHE_MIN_DELAY_MINS
+                && cacheDelayMins < Constants.CACHE_MAX_DELAY_MINS) {
+            this.cacheDelayMins = cacheDelayMins;
```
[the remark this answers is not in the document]
Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/09b48b314fa878b08f5287ee7ab2026ac5f29a01

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
isudana: So you are getting this warning all the time? :)

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,282 @@
...
+            this.cacheSize = cacheAllocatedSize;
+        }
+        log.warn("The cache size is out of range. Hence, using the default cache size value of "
+                + Constants.CACHE_DEFAULT_ALLOCATED_SIZE + ".");
+        if (cacheDelayMins != null && cacheDelayMins > Constants.CACHE_MIN_DELAY_MINS
+                && cacheDelayMins < Constants.CACHE_MAX_DELAY_MINS) {
+            this.cacheDelayMins = cacheDelayMins;
+        }
+        log.warn("The cache delay is out of range. Hence, using the default cache delay value of "
+                + Constants.CACHE_DEFAULT_DELAY_MINS + ".");
```
isudana, line 71: same as above, this will get printed all the time.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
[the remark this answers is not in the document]
isudana: this warn get printed all the time

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
isudana: then we shouldnt print a info log too.

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/d02fc9094da18bb166796929d8009da570dfb0b2

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2119/commits/d02fc9094da18bb166796929d8009da570dfb0b2

isudana (approved):

PR #2122, opened 2023-12-13
Add new attributes to ConnectionsView bean

thuva9872:
## Purpose
Synapse Diagnostic tool needs the total count of request received in the last minute. So this PR adds new attributes to the existing ConnectionView bean.


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/jmx/ConnectionsView.java
```diff
@@ -53,10 +53,13 @@ public class ConnectionsView implements ConnectionsViewMBean {
     private static final int LONG_DATA_COLLECTION_PERIOD = 60 * 5;
 
     private static final int SAMPLES_PER_HOUR = (60 * 60)/LONG_DATA_COLLECTION_PERIOD;
-
+    private Queue<Integer> lastSecondRequestQueue = new LinkedList<Integer>();
+    private  Queue<Integer> openedConnectionsDataQueue = new LinkedList<Integer>();
```
arunans23, line 57:
```suggestion
    private Queue<Integer> openedConnectionsDataQueue = new LinkedList<Integer>();
```

arunans23 (approved):

PR #2143, opened 2024-02-14
Bump okhttp okio and opentel versions to match versions in apimgt

Lakith-Rambukkanage:
## Purpose
Fixing : https://github.com/wso2/api-manager/issues/2427


Lakith-Rambukkanage:
Closing since PR is no longer valid

PR #2079, opened 2023-06-20
API Log Filtering based on API Resource Feature with Resource Matching

chamikasudusinghe:
Issue: https://github.com/wso2/api-manager/issues/1564

Refactored the methods to be accessible

PR #2150, opened 2024-02-22
Add a new method to identify the invoked API

tharikaGitHub:
## Purpose
This PR adds a new method to identify the invoked API from the collection of all APIs.

The initial change was https://github.com/wso2/wso2-synapse/pull/2083. But as this was a major refactoring, we have ported the changes from https://github.com/wso2/wso2-synapse/pull/2079 only.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
@@ -194,7 +194,6 @@ public class ApiUtils {
     }
 
     public static API getSelectedAPI(MessageContext synCtx) {
-        API selectedApi;
         //getting the API collection from the synapse configuration to find the invoked API
         Collection<API> apiSet = synCtx.getEnvironment().getSynapseConfiguration().getAPIs();
         //Since swapping elements are not possible with sets, Collection is converted to a List
@@ -204,8 +203,7 @@ public class ApiUtils {
         //identiy the api using canProcess method
         for (API api : duplicateApiSet) {
             if (identifySelectedAPI(api, synCtx, defaultStrategyApiSet)) {
-                selectedApi = api;
-                return selectedApi;
+                return api;
             }
         }
         for (API api : defaultStrategyApiSet) {
@@ -214,8 +212,7 @@ public class ApiUtils {
                 if (log.isDebugEnabled()) {
                     log.debug("Located specific API: " + api.getName() + " for processing message");
                 }
-                selectedApi = api;
-                return selectedApi;
+                return api;
             }
         }
         return null;
--- a/modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
+++ b/modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
@@ -25,6 +25,9 @@ import org.apache.http.HttpStatus;
 import org.apache.synapse.MessageContext;
 import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
+import org.apache.synapse.api.version.ContextVersionStrategy;
+import org.apache.synapse.api.version.DefaultStrategy;
+import org.apache.synapse.api.version.URLBasedVersionStrategy;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +165,94 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+            if (isBound(r, synCtx) && r.canProcess(synCtx)) {
+                acceptableResources.add(r);
+            }
+        }
+        return acceptableResources;
+    }
+
+    public static API getSelectedAPI(MessageContext synCtx) {
+        API selectedApi;
+        API defaultAPI = null;
```
isudana, line 198: defaultAPI is always null isn't it?

tharikaGitHub: Yes that's correct default API is always null in this method. However this has been taken from the logic available in the AbstractApiHandler. https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/api/AbstractApiHandler.java#L47

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +165,89 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+        for (Resource r : resources.values()) {
+            if (isBound(r, synCtx) && r.canProcess(synCtx)) {
+                acceptableResources.add(r);
+            }
+        }
+        return acceptableResources;
+    }
+
+    public static API getSelectedAPI(MessageContext synCtx) {
+        API selectedApi;
```
isudana, line 197: We can simply remove this variable.

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +165,89 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+        Set<Resource> acceptableResources = new LinkedHashSet<Resource>();
+        for (Resource r : resources.values()) {
+            if (isBound(r, synCtx) && r.canProcess(synCtx)) {
+                acceptableResources.add(r);
+            }
+        }
+        return acceptableResources;
+    }
+
+    public static API getSelectedAPI(MessageContext synCtx) {
```
isudana, line 196: add javadoc comment on what we are doing with this utility method.

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +165,89 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+                    log.debug("Located specific API: " + api.getName() + " for processing message");
+                }
+                selectedApi = api;
+                return selectedApi;
+            }
+        }
+        return null;
+    }
+
+    public static boolean identifySelectedAPI(API api, MessageContext synCtx, List defaultStrategyApiSet) {
```
isudana, line 224: do we have to make this public? what's the intension of this?

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
```diff
@@ -158,6 +165,94 @@ public static String getSubRequestPath(MessageContext synCtx) {
...
+            api.setLogSetterValue();
+            if (api.canProcess(synCtx)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Located specific API: " + api.getName() + " for processing message");
+                }
+                selectedApi = api;
+                return selectedApi;
+            }
+        }
+        if (defaultAPI != null && defaultAPI.canProcess(synCtx)) {
```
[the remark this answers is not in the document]
tharikaGitHub: no it is not getting executed. As discussed, will remove it.

modules/core/src/main/java/org/apache/synapse/api/ApiUtils.java
tharikaGitHub: Yes we can make it private

isudana (approved):

PR #2152, opened 2024-02-26
Improve cert revocation validation

Akila94:
## Purpose
- This improvement contains the certificate revocation validation improvement so the validation can be done without the full certificate chain sent in the request.

- The validation can also be done when only one intermediate client certificate is sent in the request. For this to happen the immediate issuer of that certificate should contain in the client trust store.

- Also a cert cache (LRU) is implemented to keep the issuer certificate found after traversing the client trust store.

- One of the following configurations can be used to enable this improvement,

1. Enforce glocally using axis2.xml. The following configuration should be added inside the ```<transportReceiver name="https" class="org.apache.synapse.transport.passthru.PassThroughHttpMultiSSLListener">```.

```
<parameter name="CertificateRevocationVerifier" enable="true">
    <CacheSize>50</CacheSize>
    <CacheDelay>15</CacheDelay>
    <FullChainValidation>false</FullChainValidation>
    <ExpiryValidation>true</ExpiryValidation>
</parameter>
```

2. Enforce per security listener profile. The following configuration should be added inside the relevant profile.

```
<CertificateRevocationVerifier>
    <Enable>true</Enable>
    <CacheSize>50</CacheSize>
    <CacheDelay>15</CacheDelay>
    <FullChainValidation>false</FullChainValidation>
    <ExpiryValidation>true</ExpiryValidation>
</CertificateRevocationVerifier>
```

This also fixes the test failures that occurred in the issue: https://github.com/wso2/wso2-synapse/pull/2146


modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/certificatevalidation/CertificateVerificationManager.java
```diff
@@ -0,0 +1,304 @@
...
+            this.cacheDelayMins = cacheDelayMins;
+        } else {
+            log.warn("The cache delay is out of range. Hence, using the default cache delay value of "
+                    + Constants.CACHE_DEFAULT_DELAY_MINS + ".");
+        }
+
+        this.isFullCertChainValidationEnabled = isFullCertChainValidationEnabled;
+        this.isCertExpiryValidationEnabled = isCertExpiryValidationEnabled;
+    }
+
```
arunans23, line 98:
```suggestion
    private boolean isFullCertChainValidationEnabled = true;
    private boolean isCertExpiryValidationEnabled = false;
    private static final Log log = LogFactory.getLog(CertificateVerificationManager.class);

    public CertificateVerificationManager(Integer cacheAllocatedSize, Integer cacheDelayMins) {

        this(cacheAllocatedSize, cacheDelayMins, true, false);
    }

    public CertificateVerificationManager(Integer cacheAllocatedSize, Integer cacheDelayMins,
                                          boolean isFullCertChainValidationEnabled,
                                          boolean isCertExpiryValidationEnabled) {

        if (cacheAllocatedSize != null && cacheAllocatedSize > Constants.CACHE_MIN_ALLOCATED_SIZE
                && cacheAllocatedSize < Constants.CACHE_MAX_ALLOCATED_SIZE) {
            this.cacheSize = cacheAllocatedSize;
        } else {
            log.warn("The cache size is out of range. Hence, using the default cache size value of "
                    + Constants.CACHE_DEFAULT_ALLOCATED_SIZE + ".");
        }
        if (cacheDelayMins != null && cacheDelayMins > Constants.CACHE_MIN_DELAY_MINS
                && cacheDelayMins < Constants.CACHE_MAX_DELAY_MINS) {
            this.cacheDelayMins = cacheDelayMins;
        } else {
            log.warn("The cache delay is out of range. Hence, using the default cache delay value of "
                    + Constants.CACHE_DEFAULT_DELAY_MINS + ".");
        }

        this.isFullCertChainValidationEnabled = isFullCertChainValidationEnabled;
        this.isCertExpiryValidationEnabled = isCertExpiryValidationEnabled;
    }

```

Akila94: Fixed with https://github.com/wso2/wso2-synapse/pull/2152/commits/66c6de94dabe61d56c3aa5df7fce7cf67f3e2253

arunans23 (approved):

PR #2185, opened 2024-06-11
Disable default access logger and utilize log4j2 for http access logs

arunans23:
Resolves to https://github.com/wso2/micro-integrator/issues/3281 

Currently, HTTP Access logs use a custom implementation to write the logs to the file. This has a request and response queue to hold the logs and write them to the file using a Timer task. In a high-traffic scenario, this becomes a bottleneck because the Heap memory easily fills up because of the logs added to the queuea.

This commit disables the current implementation and utilizes log4j2 to write the access logs to the file system which is more efficient. Also, timer tasks are removed and LinkedBlockingQueue is used to handle logs asynchronously in a timely manner.


SanojPunchihewa:
Since the current implementation is no longer used, Shall we remove those codes?

arunans23:
> Since the current implementation is no longer used, Shall we remove those codes?

I kept the existing code as it is, incase there are users who want to publish to custom files as before. Also, APIM product might also utilize the existing behavior.

SanojPunchihewa (approved):

PR #2222, opened 2024-10-10
Introducing RequiredFileNamePatterns feature to read files depending on the existence of other files

olhepp-avintis:
## Purpose
The purpose of this pull request is to introduce a feature covering a common use case, which is reading a file depending of the existence of another file.

This feature is particularly useful in cases, where two files are bound together and should be processed nearly at the same time in a given order, once both of them exist in a given location (e.g. receipt and warranty document for the same purchase).

## Release note
Added RequiredFileNamePatterns feature to process files depending on the existence of other files using regular expressions.

### DIFF
--- a/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
+++ b/modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
@@ -64,6 +64,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


modules/transports/core/vfs/src/main/java/org/apache/synapse/transport/vfs/VFSTransportListener.java
```diff
@@ -1247,6 +1274,76 @@ public int compare(FileObject o1, FileObject o2) {
...
+            String fileName2 = removeFixedPatterns(o2, requiredFileNamePatterns);
+            return fileName1.compareTo(fileName2);
+        }
+
+        private String removeFixedPatterns(FileObject child, String requiredFileNamePatterns) {
+        	//get base name of file
+        	String baseName = child.getName().getBaseName();
+        	//separate the patterns
+        	String[] patterns = requiredFileNamePatterns.split(",");
+        	for(int i=0;i<patterns.length;i++){
```
chanikag, line 1322:
```suggestion
        	for(int i = 0; i < patterns.length; i++){
```

chanikag:
@olhepp-avintis,

Thank you for the contribution. Could you please elaborate more on the usecase with some examples to get an idea of this feature.

Thanks.

chanikag:
Hi @olhepp-avintis,

Could you please elaborate more on the usecase with some examples on how this feature works, to get an idea.

Thanks,
Chanika

chanikag:
Closing the PR, as it is not possible to proceed without having a proper understanding of the usecase.

PR #2230, opened 2024-11-06
Improve log mediator to support string templating

SanojPunchihewa:
## Purpose

In the current log mediator, the user needs to add a property and use XPATH concatenations to define a log message. With this improvement, the user can define a string template for the log message. In addition to the log message, the user can add additional properties to be logged.

```xml
<log category="INFO">
    <message>Test clone path 1 "logging" ${var.var1}</message>
</log>
```

```xml
<log category="INFO">
    <message>Test message "logging" ${var.var1}</message>
    <property name="var2" value="test-prop1"/>
    <property name="server-header" expression="${headers.Server}"/>
</log>
```

To maintain backward compatibility when the user has set a `level`, the message will be printed in addition to the content printed according to `level`

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
@@ -29,17 +29,15 @@ import org.apache.synapse.MessageContext;
 import org.apache.synapse.SynapseLog;
 import org.apache.synapse.commons.json.JsonUtil;
 import org.apache.synapse.commons.CorrelationConstants;
-import org.apache.synapse.config.xml.SynapsePath;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.mediators.AbstractMediator;
 import org.apache.synapse.mediators.MediatorProperty;
 import org.apache.synapse.util.InlineExpressionUtil;
+import org.jaxen.JaxenException;
 
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /**
  * Logs the specified message into the configured logger. The log levels specify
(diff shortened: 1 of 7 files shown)


modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
```diff
@@ -197,4 +202,48 @@ private static boolean isValidXML(String stringToValidate) {
...
+     * @param inlineText Inline text string
+     * @return true if the string contains inline synapse expressions, false otherwise
+     */
+    public static boolean isInlineSynapseExpressionsContentAware(String inlineText) {
+
+        Matcher matcher = SYNAPSE_EXPRESSION_PLACEHOLDER_PATTERN.matcher(inlineText);
+        while (matcher.find()) {
+            // Extract the expression inside ${...}
+            String placeholder = matcher.group(1);
+            if (placeholder.contains("xpath(") || placeholder.contains("payload.") || placeholder.contains("$.")) {
```
GDLMadushanka, line 220: we might have a prob here for instances like ${toUpper(payload.name)}

modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
```diff
@@ -197,4 +202,48 @@ private static boolean isValidXML(String stringToValidate) {
...
+    public static String processInLineSynapseExpressionTemplate(MessageContext synCtx, String template)
+            throws JaxenException {
+
+        Matcher matcher = SYNAPSE_EXPRESSION_PLACEHOLDER_PATTERN.matcher(template);
+        StringBuffer result = new StringBuffer();
+        while (matcher.find()) {
+            // Extract the expression inside ${...}
+            String placeholder = matcher.group(1);
+            SynapseExpression expression = new SynapseExpression(placeholder);
+            String replacement = expression.stringValueOf(synCtx);
```
GDLMadushanka, line 243: if the stringValueOf fails we get the value "unknown" should we print a warn log outside the while loop saying processed template might have an incorrect result since evaluation of some expressions failed.

modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
SanojPunchihewa: I used the same content aware check we have in the Synapse expressions here as well.

GDLMadushanka: +1

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -288,6 +302,16 @@ public void setCategory(int category) {
         }
     }
 
+    public String getMessageTemplate() {
+
```
anupama-pathirage, line 306: Why an additional new line?

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
```diff
@@ -288,6 +302,16 @@ public void setCategory(int category) {
...
+        return messageTemplate;
+    }
+
+    public void setMessageTemplate(String messageTemplate) {
+
```
anupama-pathirage, line 311: same as above

modules/core/src/main/java/org/apache/synapse/mediators/builtin/LogMediator.java
SanojPunchihewa: It's added by the `wso2_code_style` for idea.

modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
SanojPunchihewa: We cannot distinguish the `unknown` at the mediator level. We may need to print the warn log at the evaluation.

GDLMadushanka (approved):

PR #2231, opened 2024-11-06
Add Scatter Gather mediator

SanojPunchihewa:
## Purpose

Add new mediator to clone the messages and aggregate the results.

**Syntax**

```xml
<scatter-gather parallel-execution=(true | false) result-target=(body | variable) content-type=(JSON | XML) root-element=(string) >
    <aggregation expression="expression" condition="expression" timeout="long"
    min-messages="expression" max-messages="expression"/>
        <sequence>
        (mediator)+
        </sequence>+
</scatter-gather>
```

**Sample configuration**

```xml
<scatter-gather parallel-execution="true">
    <aggregation value="json-eval($)"/>
        <sequence>
            <payloadFactory media-type="json" template-type="default">
                <format>{
                    "pet": {
                        "name": "pet1",
                        "type": "dog"
                    },
                    "status": "success"
                    }</format>
                <args>
                </args>
            </payloadFactory>
            <log category="INFO" level="custom">
                <property name="Message" value="==== DONE scatter target 1 ===="/>
            </log>
        </sequence>
        <sequence>
            <call>
                <endpoint>
                    <http method="get" uri-template="http://localhost:5454/api/pet2"/>
                </endpoint>
            </call>
            <log category="INFO" level="custom">
                <property name="Message" value="==== DONE scatter target 2 ===="/>
            </log>
        </sequence>
</scatter-gather>
```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
@@ -430,6 +430,16 @@ public class SpanHandler implements OpenTelemetrySpanHandler {
         }
     }
 
+    public void handleScatterGatherFinishEvent(MessageContext messageContext) {
+        TracingScope tracingScope = tracingScopeManager.getTracingScope(messageContext);
+            synchronized (tracingScope.getSpanStore()) {
+                cleanupContinuationStateSequences(tracingScope.getSpanStore(), messageContext);
+                SpanWrapper outerLevelSpanWrapper = tracingScope.getSpanStore().getOuterLevelSpanWrapper();
+                tracingScope.getSpanStore().finishSpan(outerLevelSpanWrapper, messageContext);
+                tracingScopeManager.cleanupTracingScope(tracingScope.getTracingScopeId());
+            }
+    }
+
     @Override
     public void handleStateStackInsertion(MessageContext synCtx, String seqName, SequenceType seqType) {
         TracingScope tracingScope = tracingScopeManager.getTracingScope(synCtx);
--- a/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
+++ b/modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
@@ -432,12 +432,21 @@ public class SpanHandler implements OpenTelemetrySpanHandler {
 
     public void handleScatterGatherFinishEvent(MessageContext messageContext) {
         TracingScope tracingScope = tracingScopeManager.getTracingScope(messageContext);
-            synchronized (tracingScope.getSpanStore()) {
-                cleanupContinuationStateSequences(tracingScope.getSpanStore(), messageContext);
-                SpanWrapper outerLevelSpanWrapper = tracingScope.getSpanStore().getOuterLevelSpanWrapper();
-                tracingScope.getSpanStore().finishSpan(outerLevelSpanWrapper, messageContext);
-                tracingScopeManager.cleanupTracingScope(tracingScope.getTracingScopeId());
-            }
+        synchronized (tracingScope.getSpanStore()) {
+            cleanupContinuationStateSequences(tracingScope.getSpanStore(), messageContext);
+            cleanUpActiveSpans(tracingScope.getSpanStore(), messageContext);
+            SpanWrapper outerLevelSpanWrapper = tracingScope.getSpanStore().getOuterLevelSpanWrapper();
+            tracingScope.getSpanStore().finishSpan(outerLevelSpanWrapper, messageContext);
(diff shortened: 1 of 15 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/v2/ScatterGather.java
```diff
@@ -0,0 +1,910 @@
...
+    private long completionTimeoutMillis = 0;
+    private Value maxMessagesToComplete;
+    private Value minMessagesToComplete;
+    private SynapsePath correlateExpression = null;
+    private SynapsePath aggregationExpression = null;
+    private boolean parallelExecution = true;
+    private Integer statisticReportingIndex;
+    private String contentType;
+    private String resultTarget;
+    private MessageContext originalMessageContext;
```
chanikag, line 102: Is this thread safe?

SanojPunchihewa: Fixed with 684fb76

modules/core/src/main/java/org/apache/synapse/mediators/MediatorWorker.java
```diff
@@ -86,7 +97,25 @@ public void run() {
                 debugManager.advertiseMediationFlowStartPoint(synCtx);
             }
 
-            seq.mediate(synCtx);
+            // If this is a scatter message, then we need to use the continuation state and continue the mediation
```
GDLMadushanka, line 100: can we evaluate boolean result = seq.mediate(synCtx); without any check and modify the check to if (isScatterMessage(synCtx) && result ) {

modules/core/src/main/java/org/apache/synapse/mediators/v2/ScatterGather.java
```diff
@@ -0,0 +1,821 @@
...
+            addChildren(list, element);
+        }
+    }
+
+    private MessageContext getAggregatedMessage(Aggregate aggregate) {
+
+        MessageContext newCtx = null;
+        if (Objects.equals(contentType, JSON_TYPE)) {
+            JsonArray jsonArray = new JsonArray();
+                log.debug("Merging aggregated JSON responses to body");
```
GDLMadushanka, line 622: should we check before log debug

modules/core/src/main/java/org/apache/synapse/mediators/MediatorWorker.java
SanojPunchihewa: Fixed with 1b4379f

modules/core/src/main/java/org/apache/synapse/mediators/v2/ScatterGather.java
SanojPunchihewa: Since we are not doing any string concatenations, we don't need to check debugEnabled

modules/core/src/main/java/org/apache/synapse/config/xml/ScatterGatherMediatorFactory.java
```diff
@@ -0,0 +1,185 @@
...
+
+        OMAttribute parallelExecAttr = elem.getAttribute(PARALLEL_EXEC_Q);
+        if (parallelExecAttr != null && parallelExecAttr.getAttributeValue().equals("false")) {
+            asynchronousExe = false;
+        }
+        mediator.setParallelExecution(asynchronousExe);
+
+        OMAttribute contentTypeAttr = elem.getAttribute(CONTENT_TYPE_Q);
+        if (contentTypeAttr == null || StringUtils.isBlank(contentTypeAttr.getAttributeValue())) {
+            String msg = "The 'content-type' attribute is required for the configuration of a Scatter Gather mediator";
```
chanikag, line 85: We discussed to derive this from the message context, and make this optional right? Was that changed?

SanojPunchihewa: It was discussed for the Foreach mediator when we need to update the original collection.

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
```diff
@@ -430,6 +430,25 @@ private void handleCallbackFinishEvent(MessageContext messageContext) {
         }
     }
 
+    public void handleScatterGatherFinishEvent(MessageContext messageContext) {
+        TracingScope tracingScope = tracingScopeManager.getTracingScope(messageContext);
+            synchronized (tracingScope.getSpanStore()) {
```
GDLMadushanka, line 435: formatting

GDLMadushanka (dismissed):

modules/core/src/main/java/org/apache/synapse/aspects/flow/statistics/tracing/opentelemetry/management/handling/span/SpanHandler.java
SanojPunchihewa: Fixed with e0c7d37

GDLMadushanka (approved):

PR #2239, opened 2024-11-26
Improve connector dependency deployment

arunans23:
$subject


modules/core/src/main/java/org/apache/synapse/libraries/LibClassLoader.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.synapse.libraries;
+
+import java.net.URLClassLoader;
+import java.net.URL;
+
+public class LibClassLoader extends URLClassLoader {
+
+        public LibClassLoader(URL[] urls, ClassLoader parent) {
```
GDLMadushanka, line 27: formatting

GDLMadushanka (approved):

PR #2242, opened 2024-11-30
Handle new resource file key values. 

chathuranga-jayanath-99:
## Purpose
With the introduction of storing resource files in the resource directory, new keys prefixed with `resources:...` will be used. Since files uploaded to this structure are copied to a specific location during deployment, these new keys have to be mapped when resolving values.


modules/core/src/main/java/org/apache/synapse/config/Utils.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.synapse.config;
+
+public class Utils {
+
+    public static final String RESOURCES_IDENTIFIER = "resources:";
+    public static final String CONVERTED_RESOURCES_IDENTIFIER = "gov:mi-resources/";
+
+    public static String transformFileKey(String fileKey) {
```
arunans23, line 26: Shall we add this util method to https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/config/SynapseConfigUtils.java rather than creating a new one?

chathuranga-jayanath-99: sure

arunans23 (approved):

PR #2246, opened 2024-12-02
Modify synapse expression return object for JSON

GDLMadushanka:
Change the default error response from null to ""


modules/core/src/main/java/org/apache/synapse/config/xml/ValueFactory.java
```diff
@@ -220,6 +225,17 @@ public SynapseJsonPath createSynJsonPath(String key) {
         return synapseJsonPath;
     }
 
+    public SynapseExpression createSynapseExpression(String key) {
```
SanojPunchihewa, line 228: Shall we have this private?

GDLMadushanka: +1 fixed

PR #2250, opened 2024-12-04
Add variable support for call mediator and improve synapse expression

arunans23:
This PR adds variable support to Call mediator target so that the response elements (payload, headers and attributes) can be stored in a variable. 

```
          <call>
            <endpoint>
               <http method="POST" uri-template="https://run.mocky.io/v3/fc50337e-e54d-45a7-8bc8-f5c1b53b2519"/>
            </endpoint>
            <target type="variable">{${params.funcParams.targetVariable}}</target>
         </call>
```

The Synapse expression has been improved to handle Map type variables so that the above stored response elements can be accessed.

Eg:

```
         <log level="simple">
            <property name="messagenew" expression="${var.resultMocky['payload']}"/>
            <property name="statusCode" expression="${var.resultMocky['attributes'].statusCode}"></property>
            <property name="content-length" expression="${var.resultMocky['headers']['Content-Length']}"></property>
         </log>
```

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
@@ -42,6 +42,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
 import org.apache.synapse.SynapseLog;
 import org.apache.synapse.commons.json.Constants;
@@ -50,6 +51,7 @@ import org.apache.synapse.config.xml.SynapsePath;
 import org.apache.synapse.config.xml.XMLConfigConstants;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.mediators.eip.EIPUtils;
+import org.apache.synapse.util.CallMediatorEnrichUtil;
 import org.apache.synapse.util.InlineExpressionUtil;
 import org.apache.synapse.util.xpath.SynapseJsonPath;
 import org.apache.synapse.util.xpath.SynapseXPath;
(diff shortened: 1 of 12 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
```diff
@@ -459,6 +466,18 @@ public void insertJson(MessageContext synCtx, Object sourceJsonElement, SynapseL
                 }
                 break;
             }
+            case EnrichMediator.VARIABLE:
+                if (action.equalsIgnoreCase(ACTION_REPLACE)) {
```
SanojPunchihewa, line 470: If we get a requirement to enrich to variables in other mediator, this may lead to a conflict. WDYT?

modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
```diff
@@ -459,6 +466,18 @@ public void insertJson(MessageContext synCtx, Object sourceJsonElement, SynapseL
...
+                    Map<String, Object> result = new HashMap<>();
+                    result.put("payload", sourceJsonElement);
+                    Map transportHeaders = (Map)((Axis2MessageContext) synCtx).getAxis2MessageContext()
+                            .getProperty(org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS);
+                    JsonObject headers = EIPUtils.convertMapToJsonObj(transportHeaders);
+                    result.put("headers", headers);
+                    result.put("attributes", CallMediatorEnrichUtil.populateTransportAttributes(synCtx));
+                    synCtx.setVariable(variable.evaluateValue(synCtx), result);
```
GDLMadushanka, line 478: can variable.evaluateValue(synCtx) be null or empty ?

modules/core/src/main/java/org/apache/synapse/mediators/elementary/Target.java
arunans23: Since we are not giving priority to the Enrich mediator going forward, we don't have to consider this.

GDLMadushanka (approved):

PR #2254, opened 2024-12-06
Fix expression not get updated issue

GDLMadushanka:
Fix expression does not get evaluated again when values change.


arunans23 (approved):

modules/core/src/main/java/org/apache/synapse/util/synapse/expression/ast/PayloadAccessNode.java
```diff
@@ -90,6 +92,8 @@ public Set<Option> options() {
 
     @Override
     public ExpressionResult evaluate(EvaluationContext context, boolean isObjectValue) throws EvaluationException {
+        // Take a copy of the expression to avoid modifying the original expression
+        expression = unProcessedExpression;
```
arunans23, line 96: Shall we rename unProcessedExpression as expression and expression as localExpression/clonedExpression/currentExpression?

PR #2262, opened 2024-12-12
Deploy endpoints for HTTP Connector Connection during Local Entry Deployer

chathuranga-jayanath-99:
## Purpose
$subject

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/deployers/LocalEntryDeployer.java
+++ b/modules/core/src/main/java/org/apache/synapse/deployers/LocalEntryDeployer.java
@@ -30,8 +30,11 @@ import org.apache.synapse.config.Entry;
 import org.apache.synapse.config.xml.EntryFactory;
 import org.apache.synapse.config.xml.EntrySerializer;
 import org.apache.synapse.config.xml.MultiXMLConfigurationBuilder;
+import org.apache.synapse.config.xml.endpoints.EndpointFactory;
+import org.apache.synapse.endpoints.Endpoint;
 import org.apache.synapse.transport.dynamicconfigurations.KeyStoreReloaderHolder;
 import org.apache.synapse.transport.nhttp.config.SslSenderTrustStoreHolder;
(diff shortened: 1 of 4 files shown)


arunans23 (approved):

modules/core/src/main/java/org/apache/synapse/deployers/LocalEntryDeployer.java
```diff
@@ -102,6 +106,50 @@ public String deploySynapseArtifact(OMElement artifactConfig, String fileName,
...
+    }
+
+    private void deployHTTPEndpointForElement(OMElement documentElement, String fileName, Properties properties) {
+
+        try {
+            Endpoint ep = EndpointFactory.getEndpointFromElement(documentElement, false, properties);
+
+            //Set the car name
+            ep.setArtifactContainerName(customLogContent);
+            if (ep != null) {
```
Bhashinee, line 127: Can this `ep` become `null`? 
Doesn't the 
`Endpoint ep = EndpointFactory.getEndpointFromElement(documentElement, false, properties);`
always return an endpoint?

modules/core/src/main/java/org/apache/synapse/deployers/LocalEntryDeployer.java
```diff
@@ -102,6 +106,50 @@ public String deploySynapseArtifact(OMElement artifactConfig, String fileName,
...
+                }
+                getSynapseConfiguration().addEndpoint(ep.getName(), ep);
+                if (log.isDebugEnabled()) {
+                    log.debug("Endpoint Deployment from the http connection file : " + fileName + " : Completed");
+                }
+                log.info("Endpoint named '" + ep.getName()
+                        + "' has been deployed from the http connection file : " + fileName);
+            } else {
+                handleSynapseArtifactDeploymentError("Endpoint Deployment Failed. The artifact " +
+                        "described in the http connection file " + fileName + " has filed to describe an Endpoint");
```
Bhashinee, line 145: `has failed to describe an Endpoint`

PR #2264, opened 2024-12-13
Improve logging in Synapse Expressions

GDLMadushanka:

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/synapse/expression/visitor/ExpressionVisitor.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/synapse/expression/visitor/ExpressionVisitor.java
@@ -17,6 +17,8 @@
  */
 package org.apache.synapse.util.synapse.expression.visitor;
 
+import org.apache.commons.logging.Log;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/synapse/util/synapse/expression/visitor/ExpressionVisitor.java
```diff
@@ -135,6 +157,9 @@ public ExpressionNode visitFactor(ExpressionParser.FactorContext ctx) {
 
     @Override
     public ExpressionNode visitFunctionCall(ExpressionParser.FunctionCallContext ctx) {
+        if (log.isDebugEnabled()) {
+            log.debug("Visiting factor: " + ctx.getText());
```
SanojPunchihewa, line 161:
```suggestion
            log.debug("Visiting FunctionCall: " + ctx.getText());
```

GDLMadushanka: fixed

SanojPunchihewa (approved):

PR #2266, opened 2024-12-16
Add Trigger Error mediator

GDLMadushanka:


modules/core/src/main/java/org/apache/synapse/config/xml/TriggerErrorMediatorFactory.java
```diff
@@ -0,0 +1,76 @@
...
+import org.jaxen.JaxenException;
+
+import javax.xml.namespace.QName;
+import java.util.Properties;
+
+/**
+ * Creates a trigger error mediator through the supplied XML configuration
+ * <p/>
+ * <pre>
+ * &lt;triggererror type="string" (errorMessage="string" | expression="expression")/&gt;
```
SanojPunchihewa, line 35: Instead of having two attributes for literal and expression, is it possible to have a [Value](https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/config/xml/ValueFactory.java#L39) ?

modules/core/src/main/java/org/apache/synapse/config/xml/TriggerErrorMediatorFactory.java
```diff
@@ -0,0 +1,76 @@
...
+ * </pre>
+ */
+public class TriggerErrorMediatorFactory extends AbstractMediatorFactory {
+
+    private static final QName ATT_ERROR_MSG = new QName("errorMessage");
+    private static final QName ATT_TYPE = new QName("type");
+    private static final QName TRIGGER_ERROR_Q
+            = new QName(XMLConfigConstants.SYNAPSE_NAMESPACE, "triggererror");
```
SanojPunchihewa, line 43: `trigger-error` is more readable. WDYT?

rosensilva: throw-error is much better IMO

GDLMadushanka: We have the following naming convention in mediators
<datamapper config="name" inputSchema="string" inputType="string" outputSchema="string" outputType="string"/> 
<jsontransform [schema="string"]>

rosensilva: then throwError?

GDLMadushanka: +1

modules/core/src/main/java/org/apache/synapse/config/xml/TriggerErrorMediatorFactory.java
GDLMadushanka: you mean the following syntax ?
`<throwError type="string">(value | expression)</ throwError>`

SanojPunchihewa: `<throwError type="string" errorMessage="value" />` looks better

SanojPunchihewa (approved):

GDLMadushanka:
related to https://github.com/wso2/product-micro-integrator/issues/3843

PR #2271, opened 2025-01-02
Update continuation stack usage in Scatter Gather mediator

SanojPunchihewa:
## Purpose

Update continuation stack usage in Scatter Gather mediator

Fixes: https://github.com/wso2/product-micro-integrator/issues/3857

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/MediatorWorker.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/MediatorWorker.java
@@ -26,17 +26,14 @@ import org.apache.synapse.Mediator;
 import org.apache.synapse.MessageContext;
 import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.SynapseException;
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/v2/Utils.java
```diff
@@ -0,0 +1,50 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.apache.synapse.mediators.v2;
+
+import org.apache.synapse.MessageContext;
+import org.apache.synapse.SynapseConstants;
+
+public class Utils {
```
GDLMadushanka, line 24: shall we rename to a name like foreachUtils since we are not inside a dedicated package and this util is not common to all v2 mediators

SanojPunchihewa: Updated with 477dfd6

GDLMadushanka (approved):

PR #2282, opened 2025-01-11
Moved synapse expression to an isolated package 

GDLMadushanka:
we can put the jar in MI libs and call using reflection. This enables
versioning the synapse expression when needed.


modules/core/src/main/java/org/apache/synapse/util/xpath/SynapseExpression.java
```diff
@@ -15,104 +15,64 @@
...
+        super(synapseExpression, org.apache.synapse.config.xml.SynapsePath.SYNAPSE_EXPRESSIONS_PATH, log);
+        String expressionClass = SynapsePropertiesLoader.loadSynapseProperties().
+                getProperty(SynapseConstants.SYNAPSE_EXPRESSION_IMPL);
+        if (!StringUtils.isEmpty(expressionClass)) {
+            if (log.isDebugEnabled()) {
+                log.debug("Trying to load the SynapseExpression implementation");
             }
-            throw new JaxenException(errorMessage.toString());
+            try {
+                Class<?> clazz = Class.forName(expressionClass);
```
SanojPunchihewa, line 50: Dynamically loading the class per each expression may have a performance impact. Shall we try to optimise this?

GDLMadushanka: +1 fixed

GDLMadushanka:
Decided not to move this out

PR #2283, opened 2025-01-14
Improve invoke mediator to support connector response model

arunans23:
Introduced a new connector response model to handle response from connectors. The invoke mediators are checked whether they are a processing a connector template and the response is processed accordingly.

Also, the CallMediatorEnrichUtil has been renamed to MediatorEnrichUtil since it is reused here.

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/template/InvokeMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/template/InvokeMediator.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License. You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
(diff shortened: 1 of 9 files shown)


modules/core/src/main/java/org/apache/synapse/mediators/template/InvokeMediator.java
```diff
@@ -290,6 +307,82 @@ private void populateParameters(MessageContext synCtx, String templateQualifiedN
...
+
+        if (!storeResponseInVariableEnabled(synCtx)) {
+            return;
+        }
+
+        org.apache.axis2.context.MessageContext axis2MessageContext =
+                ((Axis2MessageContext) synCtx).getAxis2MessageContext();
+        Object messageType = axis2MessageContext.getProperty(org.apache.axis2.Constants.Configuration.MESSAGE_TYPE);
+        Object headers = axis2MessageContext.getProperty(TRANSPORT_HEADERS);
+        Map originalTransportHeaders = (Map) headers;
```
SanojPunchihewa, line 348: Shouldn't we copy the headers map here?

arunans23: Fixed

SanojPunchihewa (approved):

PR #2287, opened 2025-01-20
Cache inline synapse expressions

SanojPunchihewa:
## Purpose

Cache inline synapse expressions

Fixes: https://github.com/wso2/product-micro-integrator/issues/3873
Fixes: https://github.com/wso2/product-micro-integrator/issues/3878

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
+++ b/modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
@@ -35,6 +35,7 @@ import org.apache.synapse.util.xpath.SynapseJsonPath;
 import org.apache.synapse.util.xpath.SynapseXPath;
 import org.jaxen.JaxenException;
 
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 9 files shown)


modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
```diff
@@ -203,42 +204,53 @@ private static boolean isValidXML(String stringToValidate) {
...
             throws JaxenException {
 
         Matcher matcher = SYNAPSE_EXPRESSION_PLACEHOLDER_PATTERN.matcher(template);
         StringBuffer result = new StringBuffer();
         while (matcher.find()) {
             // Extract the expression inside ${...}
             String placeholder = matcher.group(1);
-            SynapseExpression expression = new SynapseExpression(placeholder);
+            SynapseExpression expression = expressionCache.get(placeholder);
+            if (expression == null) {
```
GDLMadushanka, line 250: shall we add some default placeholder text ( ex: #ERROR) (in case of regex failure) or showing the expression as it is without evaluating is enough?

modules/core/src/main/java/org/apache/synapse/util/InlineExpressionUtil.java
```diff
@@ -203,42 +204,53 @@ private static boolean isValidXML(String stringToValidate) {
...
+                expression = new SynapseExpression(placeholder);
+                expressionCache.put(placeholder, expression);
+            }
             String replacement = expression.stringValueOf(synCtx);
```
GDLMadushanka, line 254: shall we calculate replacement values for the entire cached map at once. If we have the same expression used in several places, we don't have to evaluate again.

SanojPunchihewa: Fixed with c699634

GDLMadushanka (approved):

PR #2342, opened 2025-03-12
Add OpenAPI validation to REST APIs.

GDLMadushanka:
Fixes wso2/product-micro-integrator/issues/4079

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2Sender.java
+++ b/modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2Sender.java
@@ -118,11 +118,7 @@ public class Axis2Sender {
             }
             if (!responseValidated) {
                 smc.setProperty(RESTConstants.OPENAPI_VALIDATED, true);
-                try {
-                    SchemaValidationUtils.validateAPIResponse(smc, api.getOpenAPI());
-                } catch (SynapseException ex) {
-                    handleException("Schema validation failed for the response", ex, smc);
-                }
+                SchemaValidationUtils.validateAPIResponse(smc, api.getOpenAPI());
             }
         }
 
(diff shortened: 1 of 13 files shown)


modules/core/src/main/java/org/apache/synapse/util/swagger/SchemaValidationUtils.java
```diff
@@ -0,0 +1,272 @@
...
+        Optional<String> payloadObject = Optional.empty();
+        org.apache.axis2.context.MessageContext axis2Context = ((Axis2MessageContext) messageContext)
+                .getAxis2MessageContext();
+        boolean isMessageContextBuilt = isMessageContextBuilt(axis2Context);
+        if (!isMessageContextBuilt) {
+            // Build Axis2 Message.
+            try {
+                RelayUtils.buildMessage(axis2Context);
+            } catch (IOException | XMLStreamException e) {
+                logger.error(" Unable to build axis2 message");
```
abeykoon, line 122: Should not we throw this exception out?

GDLMadushanka: Here, getting the payload is optional ex. in the case of a GET request. 
Also, I belive we should't disrupt the mediation flow until we are sure of a validation error. 

We are reusing the code in [ APIM](https://github.com/wso2/carbon-apimgt/blob/5ee9d1eb160cbf5aaac6757ba58c26e3d46ee2b7/components/apimgt/org.wso2.carbon.apimgt.gateway/src/main/java/org/wso2/carbon/apimgt/gateway/handlers/security/utils/SchemaValidationUtils.java#L115)

modules/core/src/main/java/org/apache/synapse/core/axis2/Axis2Sender.java
```diff
@@ -106,6 +109,19 @@ public static void sendOn(EndpointDefinition endpoint,
      * @param smc the Synapse message context sent as the response
      */
     public static void sendBack(org.apache.synapse.MessageContext smc) {
+
+        API api = (API) smc.getProperty(RESTConstants.PROCESSED_API);
```
chanikag, line 113: Can this break when sending back in a proxy?

GDLMadushanka: no, casting a null does not throw an error. We are checking for null in the following line

chanikag (approved):

PR #2352, opened 2025-04-16
Update the server worker thread name

asithamadhushanka12:
## Purpose
Update the server worker thread naming convention to accurately reflect the currently used worker threads on a WSO2 MI instance when monitored via MBeans.

## Approach
I have updated the server worker thread name to align with the registered MBean and the corresponding worker thread name.

Fixes: https://github.com/wso2/product-micro-integrator/issues/4134


chanikag:
This is fixed with https://github.com/wso2/wso2-synapse/pull/2354/files

PR #2389, opened 2025-09-09
Add changes to PayloadFactory mediator

GDLMadushanka:
Adds a mechanism to override and enable forcefully (provided a warning)

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
+++ b/modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
@@ -37,7 +37,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.protocol.HTTP;
 import org.apache.synapse.MessageContext;
-import org.apache.synapse.SynapseConstants;
 import org.apache.synapse.commons.json.JsonUtil;
 import org.apache.synapse.config.Entry;
 import org.apache.synapse.core.axis2.Axis2MessageContext;


modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
```diff
@@ -37,6 +37,7 @@
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.protocol.HTTP;
 import org.apache.synapse.MessageContext;
+import org.apache.synapse.SynapseConstants;
```
1uffyD9: Do we need to import this here?

modules/core/src/main/java/org/apache/synapse/mediators/transform/pfutils/TemplateProcessor.java
```diff
@@ -543,5 +551,18 @@ public void readInputFactoryProperties() {
                 inputFactory.setProperty((String) entry.getKey(), entry.getValue());
```
1uffyD9, line 551: We haven't handle any exceptions here when reading from the file and setting up. Hope it handles by the `StAXUtils.loadFactoryProperties` method?

GDLMadushanka: Yes, exceptions are handled in StAXUtils. If any error, we get a null output.

modules/core/src/main/java/org/apache/synapse/mediators/transform/PayloadFactoryMediator.java
GDLMadushanka: +1 fixed

PR #2540, opened 2026-04-22
[APIM] Fixing POST/PUT/PATCH requests failing with empty body

Milanka00:
## Purpose

This PR  fixes runtime failures for empty-body requests(POST/PUT/PATCH) when a content-aware mediation flow is used with `Content-Type: application/json` header.

Previously, empty-body handling was effectively aligned to DELETE behavior, while POST/PUT (and now PATCH) could still proceed to formatter serialization and hit JSON stream write failures.

## Approach

This PR extends the existing available empty-payload logic used for DELETE to other entity-enclosing methods, instead of introducing a new separate mechanism.

- Extended empty-payload detection in DeferredMessageBuilder from `DELETE` only to `DELETE, POST, PUT, and PATCH.`

- Reused the NO_ENTITY_BODY + MESSAGE_BUILDER_INVOKED model by introducing `RelayUtils.isRequestWithoutPayload()`

- Updated `PassThroughHttpSender.sendRequestContent() `to apply the same no-body decision to POST/PUT/PATCH, so serialization is skipped when the request is confirmed as payload-less.

- Kept existing discard flow intact for no-body requests, preserving current transport behavior patterns.

- Additionally, Added an envelope-aware guard to DeferredMessageBuilder empty-payload shortcut for DELETE/POST/PUT/PATCH. This addresses the WebSub/Webhook failure path where a second build pass could override an already materialized xformValues body and lead to downstream NPEs.

## Tested Flows

- https://github.com/wso2/api-manager/issues/4826#issuecomment-4293283046

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughConstants.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/PassThroughConstants.java
@@ -241,6 +241,7 @@ public class PassThroughConstants {
     public static final String HTTP_POST = "POST";
     public static final String HTTP_DELETE = "DELETE";
     public static final String HTTP_PUT = "PUT";
+    public static final String HTTP_PATCH = "PATCH";
     public static final String HTTP_OPTIONS = "OPTIONS";
     public static final String HTTP_CONNECT = "CONNECT";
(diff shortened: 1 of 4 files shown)


Milanka00:
Closing the PR since approach has been changed

PR #2560, opened 2026-05-14
[master] Fix duplicate VFS provider registration when multiple CARs bundle the same connector dependency

### LINKED ISSUE #4901: Using the same File connector in multiple CApps causes failures when invoking file operations
### Description

Since the same File connector is present in multiple CApps, the dependent JARs are added to the class loader multiple times. As a result, invoking the connector attempts to register the same provider repeatedly, causing failures.

### Steps to Reproduce

1. Add the File connector to two CApps.
2. Deploy both CApps to the server.
3. Invoke a resource that performs a File connector operation.

### Version

4.4.0, 4.5.0

### Environment Details (with versions)

_No response_

chathuranga-jayanath-99:
## Purpose
When multiple Carbon Applications (CARs) each bundle the same File Connector dependency (e.g. commons-vfs2-sandbox), LibDeployerUtils.addDependencyToSynapseLibrary() and createSynapseLibrary() added the same-named JAR (at different temp-dir paths) to the shared LibClassLoader multiple times. StandardFileSystemManager.configurePlugins() then loaded META-INF/vfs-providers.xml twice, causing DefaultFileSystemManager.addProvider() to throw "Multiple providers registered for URL scheme 'smb2'", breaking all File Connector operations.

Fix: add a filename-based isDependencyAlreadyLoaded() guard in both methods that skips adding a JAR to the classloader if a URL with the same filename is already present. First-time additions are unaffected; duplicate same-named jars from subsequent CARs are silently skipped (logged at DEBUG level).

Also adds LibDeployerUtilsTest (8 unit tests) covering the dedup logic.

Fixes: https://github.com/wso2/product-integrator-mi/issues/4901

### DIFF
--- a/modules/core/src/main/java/org/apache/synapse/libraries/LibClassLoader.java
+++ b/modules/core/src/main/java/org/apache/synapse/libraries/LibClassLoader.java
@@ -48,7 +48,9 @@ public class LibClassLoader extends URLClassLoader {
 
         File file = new File(path);
         ArrayList urls = new ArrayList();
-        urls.add(file.toURL());
+        if (file.exists()) {
+            urls.add(file.toURL());
+        }
         File libfiles = new File(file, "lib");
         if (!addFiles(urls, libfiles)) {
             libfiles = new File(file, "Lib");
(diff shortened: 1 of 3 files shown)


SanojPunchihewa:
How about the undeployment path? Do we need to handle the same logic during undeployment as well?

chathuranga-jayanath-99:
> How about the undeployment path? Do we need to handle the same logic during undeployment as well?

No, because currently when a CApp is undeployed, the connector is also undeployed, removing the respective class loader.

SanojPunchihewa (approved):

PR #2600, opened 2026-08-14
Add v2 HTTP access logging with time-taken and correlation ID support

### LINKED ISSUE #5004: Correlating Request and Response HTTP Access Logs Improvement
### Current Limitation

MI's default HTTP entry point is the PassThrough transport (a non-blocking NIO reactor) where the request is parsed, and the response is written on two decoupled events, logged independently by two separate background threads with no shared state.

Which leads to no clear method to correlate the response log entry with its corresponding request log entry.

### Suggested Improvement

Consolidate request and response lines for better observability.

### Version

4.5.0

SanojPunchihewa:
## Purpose

Introduce v2 access log mode (access_log_v2=true) that emits a single log line per request-response pair, with time-taken in milliseconds(%D) and correlation ID(%X) pattern elements.

Fixes: Fixes: https://github.com/wso2/product-integrator-mi/issues/5004

### DIFF
--- a/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
+++ b/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
@@ -55,6 +55,17 @@ public class Access {
 
     private static LinkedBlockingQueue<HttpRequestWrapper> requestQueue;
     private static LinkedBlockingQueue<HttpResponseWrapper> responseQueue;
+    private static LinkedBlockingQueue<RequestResponsePair> combinedQueue;
+
+    public static class RequestResponsePair {
+        final HttpRequestWrapper request;
+        final HttpResponseWrapper response;
+
+        RequestResponsePair(HttpRequestWrapper req, HttpResponseWrapper res) {
+            this.request = req;
+            this.response = res;
+        }
+    }
 
     private Date date;
 
@@ -70,6 +81,7 @@ public class Access {
         Access.accessLogger = accessLogger;
         requestQueue = new LinkedBlockingQueue<HttpRequestWrapper>();
         responseQueue = new LinkedBlockingQueue<HttpResponseWrapper>();
+        combinedQueue = new LinkedBlockingQueue<RequestResponsePair>();
         logElements = createLogElements();
         logAccesses();
     }
@@ -99,13 +111,56 @@ public class Access {
     }
 
     /**
+     * Adds a correlated request+response pair to the combined queue for single-line logging.
+     *
+     * @param request  - HttpRequest
+     * @param response - HttpResponse
+     */
+    public void addCombinedAccessToQueue(HttpRequest request, HttpResponse response) {
+        HttpRequestWrapper requestWrapper = new HttpRequestWrapper();
+        requestWrapper.setHttpRequest(request);
+        Object reqTimeMs = request.getParams().getParameter("http.request.time.ms");
+        if (reqTimeMs instanceof Long) {
+            requestWrapper.setDate(new Date((Long) reqTimeMs));
+        } else {
+            requestWrapper.setDate(new Date(AccessTimeUtil.getDate().getTime()));
+        }
+
+        HttpResponseWrapper responseWrapper = null;
+        if (response != null) {
+            long responseTimeMs = System.currentTimeMillis();
+            response.getParams().setParameter("http.response.time.ms", responseTimeMs);
+            responseWrapper = new HttpResponseWrapper();
+            responseWrapper.setHttpResponse(response);
+            responseWrapper.setDate(new Date(responseTimeMs));
+        }
+
+        combinedQueue.add(new RequestResponsePair(requestWrapper, responseWrapper));
+    }
+
+    /**
      * logs the request and response accesses.
      */
     public void logAccesses() {
         Thread logRequests = new LogRequests();
         Thread logResponses = new LogResponses();
+        Thread logCombined = new LogCombined();
         logRequests.start();
         logResponses.start();
+        logCombined.start();
+    }
+
+    private class LogCombined extends Thread {
+        public void run() {
+            while (true) {
+                try {
+                    RequestResponsePair pair = combinedQueue.take();
+                    logCombined(pair.request, pair.response);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }
     }
 
     private class LogRequests extends Thread {
@@ -155,6 +210,26 @@ public class Access {
         log.debug(logString);      //log to the console
         if (accessLogger.isLoggingEnabled) {
             accessLogger.log(logString);      //log to the file
+        }
+    }
+
+    /**
+     * Logs request and response as a single combined log line.
+     *
+     * @param request  - HttpRequestWrapper
+     * @param response - HttpResponseWrapper
+     */
+    public void logCombined(HttpRequestWrapper request, HttpResponseWrapper response) {
+        Date actualDateOfOperation = request.getDate();
+        StringBuilder result = new StringBuilder(128);
+        HttpResponse httpResponse = response != null ? response.getHttpResponse() : null;
+        for (AccessLogElement logElement : logElements) {
+            logElement.addElement(result, actualDateOfOperation, request.getHttpRequest(), httpResponse);
+        }
+        String logString = result.toString();
+        accesslog.info(logString);
+        if (accessLogger.isLoggingEnabled) {
+            accessLogger.log(logString);
         }
     }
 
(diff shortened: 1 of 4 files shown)


arunans23:
There are scenarios where MI sends two responses for one request. https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceRequest.java#L114

Do we also need to think about them? Maybe we can skip pairing 100 response code and pair only 200 and above status codes. WDYT?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -98,14 +110,57 @@ public void addAccessToQueue(HttpResponse response) {
...
+        combinedQueue.add(new RequestResponsePair(requestWrapper, responseWrapper));
+    }
+
     /**
      * logs the request and response accesses.
      */
     public void logAccesses() {
         Thread logRequests = new LogRequests();
         Thread logResponses = new LogResponses();
+        Thread logCombined = new LogCombined();
```
arunans23: Shall we create this only if the v2 config is enabled?

arunans23: Shall we also give a name to this thread so that we can easily identify it in thread dumps?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -98,14 +110,57 @@ public void addAccessToQueue(HttpResponse response) {
...
+    }
+
+    private class LogCombined extends Thread {
+        public void run() {
+            while (true) {
+                try {
+                    RequestResponsePair pair = combinedQueue.take();
+                    logCombined(pair.request, pair.response);
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
```
arunans23, line 165: If the thread gets interrupted here, the queue might not get cleared here afterwards. Do we need to think about it?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -98,14 +110,57 @@ public void addAccessToQueue(HttpResponse response) {
...
+    /**
+     * Adds a correlated request+response pair to the combined queue for single-line logging.
+     *
+     * @param request  - HttpRequest
+     * @param response - HttpResponse
+     */
+    public void addCombinedAccessToQueue(HttpRequest request, HttpResponse response) {
+        HttpRequestWrapper requestWrapper = new HttpRequestWrapper();
+        requestWrapper.setHttpRequest(request);
+        Object reqTimeMs = request.getParams().getParameter("http.request.time.ms");
```
arunans23: Shall we user a constant for "http.request.time.ms" and "http.response.time.ms"?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -666,6 +741,43 @@ public void addElement(StringBuilder buf, Date date, HttpRequest request,
         }
     }
 
+    /**
+     * write the correlation ID - %I
```
arunans23: As per tomcat docs, %I is for thread name. https://tomcat.apache.org/tomcat-9.0-doc/config/valve.html

Even if this is not related to tomcat, shall we use a different pattern for this for better readability?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessConstants.java
```diff
@@ -98,4 +100,8 @@ public static String getDirectory() {
         return AccessConfiguration.getInstance().getStringProperty(CONFIG_DIRECTORY, DIRECTORY);
     }
 
+    public static boolean isV2LoggingEnabled() {
+        return AccessConfiguration.getInstance().getBooleanProperty(CONFIG_V2_LOGGING, Boolean.FALSE);
```
arunans23: Instead of reading the boolean property everytime, shall we set class level flag at the init and check that?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -666,6 +741,43 @@ public void addElement(StringBuilder buf, Date date, HttpRequest request,
...
+     */
+    protected static class CorrelationIdElement implements AccessLogElement {
+        public void addElement(StringBuilder buf, Date date, HttpRequest request,
+                               HttpResponse response) {
+            if (request != null) {
+                Object correlationId = request.getParams().getParameter(
```
arunans23, line 756: If a user sends a correlation ID with a space by mistake, the log line will now have an extra space. Should we handle that?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
```diff
@@ -98,14 +110,57 @@ public void addAccessToQueue(HttpResponse response) {
...
+     *
+     * @param request  - HttpRequest
+     * @param response - HttpResponse
+     */
+    public void addCombinedAccessToQueue(HttpRequest request, HttpResponse response) {
+        HttpRequestWrapper requestWrapper = new HttpRequestWrapper();
+        requestWrapper.setHttpRequest(request);
+        Object reqTimeMs = request.getParams().getParameter("http.request.time.ms");
+        if (reqTimeMs instanceof Long) {
+            requestWrapper.setDate(new Date((Long) reqTimeMs));
```
isudana: When is reqTimeMs is not a long?

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/LoggingNHttpClientConnection.java
```diff
@@ -69,6 +72,7 @@ public class LoggingNHttpClientConnection extends DefaultNHttpClientConnection
     private static final Properties props = MiscellaneousUtil.loadProperties(PROPERTY_FILE);
     private static final boolean REPLAY_TXN_ENABLED = Boolean.parseBoolean(
             MiscellaneousUtil.getProperty(props, PassThroughConstants.REPLAY_TXN_ENABLED_KEY, "false"));
+    private final AtomicReference<HttpRequest> pendingAccessLogRequest = new AtomicReference<>();
```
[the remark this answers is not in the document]
SanojPunchihewa: TargetHandler processes one request at a time per connection (connection is released back to the pool only after RESPONSE_DONE), so AtomicReference is sufficient — HTTP pipelining cannot occur here.

SanojPunchihewa:
> There are scenarios where MI sends two responses for one request. https://github.com/wso2/wso2-synapse/blob/master/modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/passthru/SourceRequest.java#L114
> 
> Do we also need to think about them? Maybe we can skip pairing 100 response code and pair only 200 and above status codes. WDYT?

Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Used a bounded queue to avoid OOM issues. Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Used %X. Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/AccessConstants.java
SanojPunchihewa: Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/access/Access.java
SanojPunchihewa: Fixed with 47aed11

modules/transports/core/nhttp/src/main/java/org/apache/synapse/transport/http/conn/LoggingNHttpClientConnection.java
```diff
@@ -442,7 +465,14 @@ public HttpResponse parse() throws IOException, HttpException {
                             remote.getAddress().getHostAddress());
                 }
 
-                AccessHandler.getAccess().addAccessToQueue(message);
+                if (AccessConstants.isV2LoggingEnabled()) {
+                    HttpRequest request = pendingAccessLogRequest.getAndSet(null);
+                    if (request != null) {
+                        AccessHandler.getAccess().addCombinedAccessToQueue(request, response);
+                    }
```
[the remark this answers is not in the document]
SanojPunchihewa: Fixed with 519bb1a

arunans23 (approved):

PR #2608, opened 2026-09-10
Fix Content-Type header duplication in response issue

### LINKED ISSUE #5217: Duplicate Content-Type headers are returned with different cases
### Description

When a backend returns the `Content-Type` response header using different cases (Ex: `content-type`), the Gateway treats it as a separate header and adds another `Content-Type` header to the outgoing response.

For example, the backend returns:

`content-type: application/json`

The Gateway then adds:

`Content-Type: application/json; charset=UTF-8`

Since HTTP header names are case-insensitive, the outgoing response contains duplicate `Content-Type` headers. This may cause clients to reject or incorrectly process the response.

### Steps to Reproduce

1. Configure a backend service to return the following response header: `content-type: application/json`
2. Create an API in the Publisher Portal using this service as the backend.
3. Deploy a revision and publish the API.
4. Subscribe to the API and generate an access token.
5. Invoke the API through the Gateway:
```shell
curl -ski "https://<host>/<context>/<version>/<resource>" \
  -H "Authorization: Bearer <ACCESS_TOKEN>"
```
6. Inspect the response headers.

### Version

4.6.0

### Environment Details (with versions)

_No response_

ShakyaPr:
### Purpose

- Fixes duplicate `Content-Type` response headers when a backend returns the header using different letter casing from the pass-through transport (Ex: `content-type`).
- Fixes: https://github.com/wso2/api-manager/issues/5217

### Root Cause

- For a content-aware flow, `PassThroughHttpSender#setContentType` removes the existing `Content-Type` header before adding the value derived from the message context or formatter.
- The existing removal is an exact, case-sensitive lookup. Therefore, a backend header such as `content-type` is not removed when the transport looks for `Content-Type`. Adding the transport-generated header then creates a second entry with different casing.

### Fix

- Adds `SourceResponse#removeHeaderIgnoreCase` to remove all occurrences of a header name regardless of casing.
- Introduces the optional `http.response.duplicate.content.type.header.removal` pass-through transport configuration.
- When `Content-Type` is configured, `PassThroughHttpSender` removes any existing case variants before adding the transport-generated `Content-Type` value.
- Keeps the existing exact-match behavior when the configuration is absent or empty, preserving backward compatibility.
- Leaves the existing `SourceResponse#removeHeader` behavior unchanged.

### Configuration

```toml
[passthru_http]
"http.response.duplicate.content.type.header.removal" = true
```

### impact

- The fix is enabled by default so if anyone wants to preserve the old behavior, needs to use flag `"http.response.duplicate.content.type.header.removal" = false`
- With the configuration enabled, a backend response containing `content-type`, `CONTENT-TYPE`, or another case variant is emitted with a single transport-generated `Content-Type` header.
- Gateway-generated MCP responses continue to receive their expected `Content-Type` header.


ShakyaPr:
@Coderabbitai please approve if the review is completed and no comments

tharindu1st (dismissed):

ShakyaPr:
@Coderabbitai please approve if the review is completed and no comments

PR #2610, opened 2026-09-16
Skip building unused script engine pool for inline script mediators

Shamly-Shanawaz:
## Purpose

`ScriptMediator.initScriptEngine()` unconditionally pre-allocates a pool of `poolSize` (default 15) `ScriptEngineWrapper` instances for every script mediator.

The pool is only consumed through `getNewScriptEngine()`, which is reachable only from `mediateWithExternalScript()` (`key != null`). Inline `<script>` mediators are mediated by `mediateForInlineScript()`, which never touches the pool — so for every inline script mediator those engines stay idle for the lifetime of the mediator.

The cost is paid per deployed artifact at deployment time, so it scales with configuration size rather than with load and is fully present on an idle server.

## Approach

Guard the pool construction with `key != null`, so it is built only for external/key-referenced script mediators.

No behaviour change otherwise:
- External script pooling, `poolSize` resolution, `poll()`/`offer()` semantics and the fallback in `getNewScriptEngine()` are untouched.
- Inline mediation never read the pool before and does not now.

## Result

Measured with the equivalent change on the corresponding support branch (API Manager 4.2.0, 1,000 deployed APIs each with one inline `<script language="js">` mediator, live heap after a full GC):

| Metric | Before | After |
|---|---:|---:|
| Live heap | 2.25 GiB | 0.60 GiB |
| Script engine footprint | 780.6 MB | 128.7 MB |
| `ScriptEngineWrapper` instances | 15,000 | 0 |

`ScriptMediator` (1,000) and `SynapseXPath` (10,092) counts were unchanged, confirming identical topology.

Existing `modules/extensions` tests pass (145 tests, 0 failures).

Related Issue : https://github.com/wso2/api-manager/issues/5218


arunans23 (approved):

Shamly-Shanawaz:
@coderabbitai review

Shamly-Shanawaz:
@coderabbitai full review

